PR #2, opened 2022-04-02
Endpoints

Flavio0834:
### Description
Started creating endpoints for users and associations.
Endpoints linked to empty async functions.


.github/workflows/lintandtest.yml
```diff
@@ -30,10 +30,5 @@ jobs:
       - name: Format with black
         uses: psf/black@stable
 
-      - name: Sort imports with isort
-        uses: jamescurtin/isort-action@master
-        with:
-          requirementsFiles: "requirements.txt"
-
```
AtomHare, line 37: We should keep this

app/endpoints/amap.py
```diff
@@ -0,0 +1,117 @@
+from fastapi import APIRouter
+
+router = APIRouter()
+
+
+@router.get("/products")
```
AtomHare, line 6: we need to have `amap` on all those requests as the endpoint
```suggestion
@router.get("/amap/products")
```

app/endpoints/bdecalendar.py
```diff
@@ -0,0 +1,39 @@
+from fastapi import APIRouter
+
+router = APIRouter()
+
+
+@router.get("/events")
```
AtomHare, line 6:
```suggestion
@router.get("/calendar/events")
```
like amap

AtomHare (changes requested):
Pretty good, 
You just need to edit the routes and I'm not sure about the naming of `bdebooking.py` and `bdecalendar.py`

.github/workflows/lintandtest.yml
Flavio0834: done

app/endpoints/amap.py
Flavio0834: Adding prefix in include_router in api.py is better but same effect. Done

app/endpoints/bdecalendar.py
Flavio0834: same answer

Victor-Angot (approved):

Jhobahtes (approved):
LGTM 👍

AtomHare (approved):
LGTM

PR #8, opened 2022-04-10
Accounts management, authentication and authorization

Jhobahtes:
## Account management
 * Create, recover, update users
 * Email confirmation
## Group management
 * Create, modify
 * Default groups and account types
## Security
 * Bcrypt hash and verification
## Utils
 * SMTP email sender
## Authorization
 * OAuth password flow using JWT
 * Groups dependencies using JWT
## Tests
 * Async and database support
 > Tests can now be used and created
## Workflow
 * Python *basic* typechecking
 * Add a `.env` dotfile

## TODO
Use an email and an UUID type
Change cruds parameters and return types
Add loggin


app/core/security.py
```diff
@@ -0,0 +1,25 @@
+import secrets
+
+
+def password_validator(password: str) -> str:
+    """
+    Check the password strength, validity and remove trailing spaces.
+    This function is intended to be used as a Pydantic validator:
+    https://pydantic-docs.helpmanual.io/usage/validators/#reuse-validators
+    """
+    # TODO
```
Victor-Angot, line 10: Remove this comment

Victor-Angot (dismissed):

Victor-Angot (approved):

app/core/security.py
AtomHare: I think we should keep the TODO so we remember to add more checks like at least a number etc

PR #15, opened 2022-04-28
Workflow improvements

Jhobahtes:
### Mypy
Introduce [mypy](https://mypy.readthedocs.io/en/stable/) type checker.
Fix errors found by mypy.

### VS Code
 - sort import on save
 - replace pyright type checker by [mypy](https://mypy.readthedocs.io/en/stable/)

### GitHub workflow
 - replace [isort/isort-action@master](https://github.com/isort/isort-action) action by `isort . --check-only --diff` command. The previous one was installing a new and different version of isort (10s shorter)
 - change black action configuration. We don't want black to make modification as changes were never committed by the workflow. 
 - caching: pytest and mypy cache folder are now cached by actions/cache@v3.


AtomHare (approved):
LGTM, nice cleanup 👍

PR #46, opened 2022-08-13
Core improvements: security.txt, validation exception handler, CORS middleware and doc

Jhobahtes:
# Core improvements

This pull request add a few improvements to the core. This should be one of the last *big pull request* modifying multiple parts of Hyperion. It does not include breaking changes.

### security.txt
Based on [securitytxt.org](https://securitytxt.org/) proposal, add a `security.txt` file explaining how to report security concern about Hyperion.

### `/information` endpoint
`/information` is intended to be used as a health check endpoint. The app can access it to make sure Hyperion is available and working. It should raise a warning if this endpoint is unreachable.

It also return three information:
```python
    return schemas_core.CoreInformation(
        ready=True,
        version=settings.HYPERION_VERSION,
        minimal_titan_version=settings.MINIMAL_TITAN_VERSION,
    )
```
 - `ready` is not currently used by Hyperion. It can be changed to false when the API running but unavailable, for example on startup, or after a fatal error.
 - `version` allows client to check Hyperion version
 - `minimal_titan_version` is specific to Hyperion and should be check against the app version when the app is opened. If the version is to low, Titan should invite the user to update the app and could become unusable. 

### `validation_exception_handler()`
Using [Use the RequestValidationError body](https://fastapi.tiangolo.com/tutorial/handling-errors/#use-the-requestvalidationerror-body) documentation, catch validation exceptions which are raised by FastAPI when the API receive invalid data. Currently information logged about these issues are minimal, the handler tries to be more verbose, using *hyperion.error* logger.

### CORS middleware
Allowed CORS origins can be set from the dotenv

### `REFRESH_TOKEN_EXPIRE_MINUTES`
Set the refresh tokens duration to 14 days. This parameter may need to be modified. It should be small to limit the impact of a token theft but should be long enough to prevent user from needing to sign in every times they open the app. 

### Documentation
Make small improvements to the documentation.

### Checklist

- [x] All tests passing


benjamindenise:
Please explain objectives of this PR.

assets/security.txt
```diff
@@ -0,0 +1,5 @@
+Contact: mailto:security@myecl.fr
```
benjamindenise, line 1: Don't forget to create email.

benjamindenise (approved):

Jake-Ransom (approved):

maximeroucher (approved):

PR #60, opened 2022-09-10
Rate-limit

Jake-Ransom:
### Description

Implementation of a simple rate limiter (fixed window)

### Checklist

- [x] All tests passing


Victor-Angot (approved):

app/dependencies.py
```diff
@@ -22,13 +23,30 @@ async def get_users(db: AsyncSession = Depends(get_db)):
...
 from app.schemas import schemas_auth
+from app.utils.redis import connect
 from app.utils.tools import is_user_member_of_an_allowed_group
 from app.utils.types.groups_type import GroupType
 from app.utils.types.scopes_type import ScopeType
 
 # We could maybe use hyperion.security
 hyperion_access_logger = logging.getLogger("hyperion.access")
 
+redis_client = None  # Create a global variable for the redis client, so that it can be instancied in the startup and shutdown events
```
Jhobahtes, line 34: Could you type `redis_client`?

```suggestion
redis_client: redis.Redis | bool | None = None  # Create a global variable for the redis client, so that it can be instancied in the startup and shutdown events
```

app/dependencies.py
```diff
@@ -22,13 +23,30 @@ async def get_users(db: AsyncSession = Depends(get_db)):
...
+# Is None if the redis client is not instancied, is False if the redis client is instancied but not connected, is a redis.Redis object if the redis client is connected
+
+
+def get_redis_client(
+    settings=None,
+) -> redis.Redis | None | bool:  # settings can be None if the redis client is already instancied, so that we don't need to pass the settings to the function
+    """
+    Dependency that return the redis client
+    """
```
Jhobahtes, line 43: Putting the comment in the docstring would make it easier to see for people using the dependency

```suggestion
) -> redis.Redis | None | bool:
    """
    Dependency that return the redis client
    
    Settings can be None if the redis client is already instanced, so that we don't need to pass the settings to the function.

    Is None if the redis client is not instancied, is False if the redis client is instancied but not connected, is a redis.Redis object if the redis client is connected
    """
```

app/dependencies.py
```diff
@@ -22,13 +23,30 @@ async def get_users(db: AsyncSession = Depends(get_db)):
...
 
 # We could maybe use hyperion.security
 hyperion_access_logger = logging.getLogger("hyperion.access")
 
+redis_client = None  # Create a global variable for the redis client, so that it can be instancied in the startup and shutdown events
+# Is None if the redis client is not instancied, is False if the redis client is instancied but not connected, is a redis.Redis object if the redis client is connected
+
+
+def get_redis_client(
+    settings=None,
```
Jhobahtes, line 39:
```suggestion
    settings: Settings | None = None,
```

app/main.py
```diff
@@ -36,28 +42,53 @@ async def logging_middleware(request: Request, call_next):
...
+        client_address = f"{ip_address}:{port}"
     else:
         client_address = "unknown"
 
-    hyperion_access_logger.info(
-        f'{client_address} - "{request.method} {request.url.path}" {response.status_code} ({request_id})'
+    settings = app.dependency_overrides.get(get_settings, get_settings)()
+    redis_client = app.dependency_overrides.get(get_redis_client, get_redis_client)(
+        settings=settings
     )
```
Jhobahtes, line 55:
```suggestion
    settings: Settings = app.dependency_overrides.get(get_settings, get_settings)()
    redis_client: Redis | bool | None = app.dependency_overrides.get(get_redis_client, get_redis_client)(
        settings=settings
    )
```

app/utils/redis.py
```diff
@@ -0,0 +1,35 @@
...
+from app.core.config import Settings
+
+
+def connect(settings: Settings) -> redis.Redis | bool:
+    try:
+        redis_client = redis.Redis(host=settings.REDIS_HOST, port=settings.REDIS_PORT)
+        redis_client.ping()  # Test the connection
+    except redis.exceptions.ConnectionError as e:
+        print("Redis connection error: ", e)
+        print("Check the Redis configuration  or the Redis server")
```
Jhobahtes, line 12: Could you use `hyperion.error` logger?

Jhobahtes (changes requested):
Impressive work! Thanks
I added a few comments, mostly about type and documentation

app/utils/redis.py
Jake-Ransom: I think that using the the logger in this file wouldn't make much sense. Then how about I to just delete the exception handling in this file and put it in the dependencies ?

Jhobahtes: I agree, it makes sense

Jhobahtes (approved):

PR #68, opened 2022-10-09
Fix a good number of typos

Pokegali:
### Description

Fixed typos in the comments


armanddidierjean (approved):
Thank you for this pull request. It's a really nice job!
I just removed a link to the "reset password" form as it is currently not implemented in order to not mislead our alpha testers.

PR #110, opened 2023-01-12
Add Minecraft auth provider

Pokegali:
### Description

Added an auth provider for the minecraft server

### DIFF
--- a/app/utils/auth/providers.py
+++ b/app/utils/auth/providers.py
@@ -228,3 +228,20 @@ class SynapseAuthClient(BaseAuthClient):
             ),
             "email": user.email,
         }
+
+
+class MinecraftAuthClient(BaseAuthClient):
+    # If no redirect_uri are hardcoded, the client will need to provide one in its request
+    redirect_uri: str | None = None
+    # Set of scopes the auth client is authorized to grant when issuing an access token.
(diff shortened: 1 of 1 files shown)


Jake-Ransom (approved):

app/utils/auth/providers.py
```diff
@@ -196,3 +196,21 @@ def get_userinfo(cls, user: models_core.CoreUser):
...
+    # Set of scopes the auth client is authorized to grant when issuing an access token.
+    # See app.utils.types.scopes_type.ScopeType for possible values
+    allowed_scopes: Set[ScopeType] = {ScopeType.profile}
+
+    @classmethod
+    def get_userinfo(cls, user: models_core.CoreUser):
+
+        return {
+            "id": user.id,
+            "nickname": user.nickname,
```
armanddidierjean, line 213: A user may not have a surname. The firstname may be used in this case

Flavio0834: maybe it would be better to use Firstname + Lastname knowing that the surname will become your in-game name

Pokegali:
After some discussion, it's okay if we get nothing in the nickname field. The minecraft username will be left unchanged if it is the case

PR #112, opened 2023-01-15
Core User creation promo and floor

armanddidierjean:
### Description

- Add a floor enumeration
- Support new email address
- Prevent users from changing their name and firstname
- Return phone numbers

A database migration may be required

### DIFF
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -5,6 +5,7 @@ from urllib.parse import parse_qs, urlparse
 from app.main import app
 from app.models import models_core
 from app.utils.examples import examples_auth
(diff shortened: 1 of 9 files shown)


tests/test_auth.py
```diff
@@ -19,7 +19,7 @@ async def startuptest():
             firstname="Antoine",
             nickname="Nickname",
             birthday=date.fromisoformat("2000-01-01"),
-            floor="M16",
+            floor="Autre",
```
Jake-Ransom, line 22: Why not FloorsType.Autre ?
```suggestion
            floor=FloorsType.Autre,
```

maximeroucher (approved):

PR #125, opened 2023-02-07
Put entity as nullable in schema_bdebooking

Jake-Ransom:


Malem38 (approved):
Thank you very much for this beautiful work !

PR #186, opened 2023-06-17
Fix ical rrule

Petitoto:
## Description

- fix RRULE when creating iCal file
- add an endpoint to allow global admin to manually create iCal file

Refs for RRULE fix : 
https://stackoverflow.com/questions/20143590/how-to-add-rrule-to-icalendar-event-in-python
https://icalendar.readthedocs.io/en/latest/_modules/icalendar/prop.html#vRecur

### DIFF
--- a/app/endpoints/calendar.py
+++ b/app/endpoints/calendar.py
@@ -229,7 +229,7 @@ async def delete_bookings_id(
 
 
 @router.post(
-    "/calendar/ical",
+    "/calendar/ical/create",
     status_code=204,
     tags=[Tags.calendar],
 )
--- a/app/endpoints/calendar.py
+++ b/app/endpoints/calendar.py
@@ -233,7 +233,7 @@ async def delete_bookings_id(
     status_code=204,
     tags=[Tags.calendar],
 )
-async def create_product(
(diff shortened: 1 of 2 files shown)


app/endpoints/calendar.py
```diff
@@ -228,6 +228,25 @@ async def delete_bookings_id(
         )
 
 
+@router.post(
+    "/calendar/ical",
+    status_code=204,
+    tags=[Tags.calendar],
+)
+async def create_product(
```
armanddidierjean, line 236: The name could be a bit more explicit, like:
```suggestion
async def recreate_ical_file(
```

app/endpoints/calendar.py
```diff
@@ -228,6 +228,25 @@ async def delete_bookings_id(
         )
 
 
+@router.post(
+    "/calendar/ical",
```
armanddidierjean, line 232: I would suggest a route like `POST /calendar/ical/recreate` or `POST /calendar/ical/create` to emphasize what the endpoint is doing.

Petitoto: /calendar/ical/create

PR #188, opened 2023-06-18
Use timezone from datetime module

armanddidierjean:
instead of pytz.

The timezone can be changed using the dotenv. 
NOTE: the chosen timezone does not have any effects on validators as we can not access settings inside a synchronous validator.

### DIFF
--- a/.env
+++ b/.env
@@ -40,7 +40,7 @@ REDIS_LIMIT = 1000
 REDIS_WINDOW = 60
 
 # Timezone #
-TIMEZONE = "Europe/Paris"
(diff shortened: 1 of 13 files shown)


Petitoto:
It would be better if Titan was able to properly handle timezones, so we wouldn't have to force the timezone from Hyperion. Here, datetime's "PTH2" does not handle summer/winter time changes.
Once Titan will send time as UTC, we won't have to artificially change the timezone.

armanddidierjean:
Superseded by #360

PR #192, opened 2023-07-01
feat: add field validation in config

armanddidierjean:
to be able to check for incorrect values before runtime and provide more explicit error messages

Requires: https://github.com/aeecleclair/Hyperion/pull/193


Petitoto:
Did not check that tests pass with #193 but lgtm

Petitoto (approved):

PR #212, opened 2023-07-17
Fix open redirect vulnerability

Petitoto:
An open redirect vulnerability allowed attackers to exploit the redirect_uri query parameter to redirect a user from the authentication portal to an external website.

This vulnerability allows advanced phishing attacks. Token stealing is already mitigated.

This PR enforce the hardcoded redirect_uri in providers settings to mitigate the vulnerability.

When merging this PR, .env file needs to be updated as follow:
`AUTH_CLIENTS =  [["client id", "client secret", "redirect_uri", "app.utils.auth.providers class name"]]`


Petitoto:
> ⚠️ GitGuardian has uncovered 4 secrets following the scan of your pull request.

False positive
It refers to .env.test

armanddidierjean (approved):
This will require all dev to update their dotenv and modification for the prod and preprod.

armanddidierjean:
@Jake-Ransom what do you think of this modification?

maximeroucher (approved):

Jake-Ransom:
Why do you want to put the redirect Uri in the .env and not juste make the "redirect_uri" parameter in the providers class mandatory ? That way, there would not be a need to update all the .env files

Otherwise that's a good catch, I agree !

Petitoto:
> Why do you want to put the redirect Uri in the .env and not juste make the "redirect_uri" parameter in the providers class mandatory ? That way, there would not be a need to update all the .env files
> 
> Otherwise that's a good catch, I agree !
> 

This is how it was done at first, but it appears that it was too dependant on ECLAIR's infrastructure, so we decided to move it in .env

Jake-Ransom (approved):

PR #236, opened 2023-08-21
Module visibility

Shaihh:
### Description

Add the representation of modules as a class. 
All new modules must now be added to ModuleList indicating to Titan which modules must be seen by which groups

### Checklist

- [x] All tests passing


app/cruds/cruds_module_visibility.py
```diff
@@ -0,0 +1,96 @@
...
+        select(
+            models_core.ModuleVisibility.allowed_group_id,
+        ).where(models_core.ModuleVisibility.root == root)
+    )
+
+    resultList = result.unique().scalars().all()
+
+    if resultList:
+        return resultList
+    else:
```
maximeroucher, line 50: You can remove the else here since the first condition returns something

app/endpoints/core.py
```diff
@@ -127,3 +133,104 @@ async def get_style_file(
...
+async def get_user_modules_visibility(
+    db: AsyncSession = Depends(get_db),
+    user: models_core.CoreUser = Depends(is_user_a_member),
+):
+    """
+    Get group user accessible root
+
+    **This endpoint is only usable by everyone**
+    """
+    result = await cruds_module_visibility.get_modules_by_user(user=user, db=db)
```
maximeroucher, line 184: Do you need a variable assignation ?

maximeroucher (approved):

app/cruds/cruds_module_visibility.py
armanddidierjean, line 1: As models are in `models_core` I think theses cruds could be in `cruds_core`

app/cruds/cruds_module_visibility.py
```diff
@@ -0,0 +1,95 @@
+from sqlite3 import IntegrityError
```
armanddidierjean, line 1: I'm not sure but would this work with Posgresql databases?

app/cruds/cruds_module_visibility.py
```diff
@@ -0,0 +1,95 @@
...
+async def delete_module_visibility(
+    root: str,
+    allowed_group_id: str,
+    db: AsyncSession,
+):
+    await db.execute(
+        delete(models_core.ModuleVisibility).where(
+            (models_core.ModuleVisibility.root == root)
+            & (models_core.ModuleVisibility.allowed_group_id == allowed_group_id)
+        )
```
armanddidierjean, line 93: `&` should work but according to [SQLAlchemy documentation](https://docs.sqlalchemy.org/en/20/tutorial/data_select.html#the-where-clause) you can provide coma separated conditions to the method:

```suggestion
        delete(models_core.ModuleVisibility).where(
            models_core.ModuleVisibility.root == root,
            models_core.ModuleVisibility.allowed_group_id == allowed_group_id
        )
```

app/endpoints/core.py
```diff
@@ -127,3 +133,103 @@ async def get_style_file(
 )
 async def get_favicon():
     return FileResponse("assets/images/favicon.ico")
+
+
+@router.get(
+    "/module_visibility/",
```
armanddidierjean, line 139: It is usually recommended to use hyphen as word separator in RestAPI url, this is also the convention we used for other Hyperion's endpoints

app/endpoints/core.py
```diff
@@ -127,3 +133,103 @@ async def get_style_file(
...
+    db: AsyncSession = Depends(get_db),
+    user: models_core.CoreUser = Depends(is_user_a_member_of(GroupType.admin)),
+):
+    """
+    Get all existing module_visibility.
+
+    **This endpoint is only usable by administrators**
+    """
+
+    returnModuleVisibilities = []
```
armanddidierjean, line 154: Prefer snake_case
```suggestion
    return_module_visibilities = []
```

app/endpoints/core.py
```diff
@@ -127,3 +133,103 @@ async def get_style_file(
...
+        module_visibility_db = models_core.ModuleVisibility(
+            root=module_visibility.root,
+            allowed_group_id=module_visibility.allowed_group_id,
+        )
+
+        return await cruds_module_visibility.create_module_visibility(
+            module_visibility=module_visibility_db, db=db
+        )
+    except ValueError as error:
+        raise HTTPException(status_code=422, detail=str(error))
```
armanddidierjean, line 221:
```suggestion
        raise HTTPException(status_code=400, detail=str(error))
```

armanddidierjean (changes requested):

armanddidierjean:
Really nice and needed functionality!

app/utils/types/module_list.py
```diff
@@ -0,0 +1,51 @@
...
+
+class ModuleList(Module, Enum):
+    advert = Module(
+        root="advert",
+        default_allowed_groups_ids=[GroupType.student, GroupType.staff],
+    )
+    admin = Module(
+        root="admin",
+        default_allowed_groups_ids=[GroupType.admin],
+    )
```
armanddidierjean, line 15: Should admin really be able to disable `admin` module?

Shaihh: Right now access to admin module is managed on Titan side but it could be better to add a special case in futur implementation. Right now it may be better to remove it from the module list as is not considered as a drawer module in Titan and does nothing when changed.

armanddidierjean (approved):
Thanks for this functionality which was really needed

PR #250, opened 2023-09-05
Fix notifications

Petitoto:
### Description

- Various fix when calling send_notification_to_user
- Fix notifications title
- Fix CustomTopic to_str() which lead to send wrongly formatted topic names to Titan
- Add get_topit_identifier to get full topic name (with identifiers) the user is subscribed to for a given topic
-  Rename get_topic_membership_XX to plural & add get_topic_memberships_by_user_id_and_topic

### Checklist

- [x] All tests passing

### DIFF
--- a/app/cruds/cruds_notification.py
+++ b/app/cruds/cruds_notification.py
@@ -241,7 +241,7 @@ async def get_topic_memberships_by_user_id_and_topic(
     user_id: str,
     custom_topic: CustomTopic,
     db: AsyncSession,
-) -> models_notification.TopicMembership | None:
+) -> Sequence[models_notification.TopicMembership]:
     result = await db.execute(
         select(models_notification.TopicMembership).where(
             models_notification.TopicMembership.user_id == user_id,
--- a/app/cruds/cruds_notification.py
+++ b/app/cruds/cruds_notification.py
@@ -246,7 +246,7 @@ async def get_topic_memberships_with_identifiers_by_user_id_and_topic(
         select(models_notification.TopicMembership).where(
             models_notification.TopicMembership.user_id == user_id,
             models_notification.TopicMembership.topic == topic,
-            models_notification.TopicMembership.topic_identifier.is_not(None),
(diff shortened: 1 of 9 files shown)


app/cruds/cruds_notification.py
```diff
@@ -237,7 +237,21 @@ async def get_topic_membership_by_user_id(
     return result.scalars().all()
 
 
-async def get_topic_membership_by_user_id_and_topic(
+async def get_topic_memberships_by_user_id_and_topic(
+    user_id: str,
+    custom_topic: CustomTopic,
```
armanddidierjean, line 242: We could pass a `topic: Topic` parameter

app/cruds/cruds_notification.py
```diff
@@ -237,7 +237,21 @@ async def get_topic_membership_by_user_id(
...
+    result = await db.execute(
+        select(models_notification.TopicMembership).where(
+            models_notification.TopicMembership.user_id == user_id,
+            models_notification.TopicMembership.topic == custom_topic.topic,
+        )
+    )
+    return result.scalars().all()
+
+
+async def get_topic_membership_by_user_id_and_full_topic(
```
armanddidierjean, line 254:
```suggestion
async def get_topic_membership_by_user_id_and_custom_topic(
```

app/endpoints/advert.py
```diff
@@ -282,7 +283,7 @@ async def create_advert(
         )
     except Exception as error:
         hyperion_error_logger.error(
-            f"Error while sending cinema recap notification, {error}"
+            f"Error while sending AMAP recap notification, {error}"
```
armanddidierjean, line 286:
```suggestion
            f"Error while sending advert notification, {error}"
```

armanddidierjean: I added this try/except, but I'm not sure whether it is really necessary, could this code really fail?

Petitoto: For now I think this try except can be useful to monitor if all is going well with notifications

PR #256, opened 2023-09-26
Campaign-authorization

landacus:
### Description

### Checklist

- [ ] All tests passing

### DIFF
--- a/app/cruds/cruds_campaign.py
+++ b/app/cruds/cruds_campaign.py
@@ -42,10 +42,11 @@ async def get_voters(db: AsyncSession) -> Sequence[models_campaign.Voters]:
 
 
 async def add_voters(
-    voters: models_campaign.Voters,
+    voters: Sequence[models_campaign.Voters],
     db: AsyncSession,
 ) -> None:
-    db.add(voters)
+    for voter in voters:
+        db.add(voter)
     try:
         await db.commit()
     except IntegrityError as err:
--- a/app/cruds/cruds_campaign.py
+++ b/app/cruds/cruds_campaign.py
@@ -41,17 +41,28 @@ async def get_voters(db: AsyncSession) -> Sequence[models_campaign.VoterGroups]:
     return result.scalars().all()
 
 
-async def add_voters(
-    voters: Sequence[models_campaign.VoterGroups],
+async def add_voter(
+    voter: models_campaign.VoterGroups,
     db: AsyncSession,
 ) -> None:
-    for voter in voters:
-        db.add(voter)
+    db.add(voter)
     try:
         await db.commit()
     except IntegrityError as err:
         await db.rollback()
         raise err
+
+
+async def delete_voter_by_group_id(
+    group_id: int,
+    db: AsyncSession,
+) -> None:
+    await db.execute(
+        delete(models_campaign.VoterGroups).where(
+            models_campaign.VoterGroups.group_id == group_id
+        )
+    )
+    await db.commit()
 
 
 async def delete_voters(
--- a/app/cruds/cruds_campaign.py
+++ b/app/cruds/cruds_campaign.py
@@ -45,8 +45,8 @@ async def get_voters_list(
     db: AsyncSession,
 ) -> list[str] | None:
     result = await db.execute(select(models_campaign.Voters))
-    voters_group = result.scalars().all()
-    if len(voters_group) == 0:
+    voters = result.scalars().all()
+    if len(voters) == 0:
         # The voters was never set in the database, we can create a default status in the database and return it
         # Since this is the only place a row can be added to the status table, there should never be more than one row in the table
         voters_model = models_campaign.Voters(group="ALL", id="id")
@@ -58,7 +58,7 @@ async def get_voters_list(
             raise ValueError(err)
 
     # The status is contained in the only result returned by the database
-    return [e.group for e in voters_group]
+    return [voter.group for voter in voters]
 
 
 async def add_voters(
@@ -81,24 +81,6 @@ async def delete_voters(
         delete(models_campaign.Voters).where(models_campaign.Voters.id == id)
     )
     await db.commit()
-
-
-async def update_voters(
-    id: str,
-    voters_update: schemas_campaign.Voters,
-    db: AsyncSession,
-) -> None:
-    """Update a campaign list."""
-    await db.execute(
-        update(models_campaign.Voters)
-        .where(models_campaign.Voters.id == id)
-        .values(**voters_update.dict(exclude_none=True))
-    )
-    try:
-        await db.commit()
-    except IntegrityError:
-        await db.rollback()
-        raise ValueError()
 
 
 async def get_voters_by_id(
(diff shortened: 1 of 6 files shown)


armanddidierjean (changes requested):
We should not merge this PR before campaigns to let us test it and make sure everything is working correctly

app/schemas/schemas_campaign.py
```diff
@@ -109,3 +109,13 @@ class Config:
...
+
+
+class Voters(BaseModel):
+    """Base schema for a vote."""
+
+    id_voters: str
+    group: str
+
+    class Config:
+        orm_mode = True
```
armanddidierjean, line 121: Could you add a docstring explaining the purpose of this schema?

app/models/models_campaign.py
```diff
@@ -69,3 +69,10 @@ class Status(Base):
 
     id: Mapped[str] = mapped_column(String, nullable=False, primary_key=True)
     status: Mapped[StatusType] = mapped_column(Enum(StatusType), nullable=False)
+
+
+class Voters(Base):
+    __tablename__ = "campaign_voters"
+
+    id: Mapped[str] = mapped_column(String, nullable=False, primary_key=True)
+    group: Mapped[str] = mapped_column(String, nullable=False)
```
armanddidierjean, line 78: Is this a group_id?
```suggestion
    group_id: Mapped[str] = mapped_column(String, nullable=False)
```

app/schemas/schemas_campaign.py
armanddidierjean: If I understood correctly, the table contains identifiers of groups that are allowed to vote

armanddidierjean: Maybe we should find a more explicit name

armanddidierjean (changes requested):

app/endpoints/campaign.py
```diff
@@ -37,13 +44,21 @@
...
+    user: models_core.CoreUser = Depends(is_user_a_member),
 ):
     """
     Return sections in the database as a list of `schemas_campaign.SectionBase`
 
-    **The user must be a member of the group AE to use this endpoint**
+    **The user must be a voter to use this endpoint**
     """
+    voters = await cruds_campaign.get_voters_list(db)
+    if voters:
```
armanddidierjean, line 55: I would be in favor of blocking every user if no allowed groups where set. This will prevent someone from opening a vote while forgetting to configure Voters groups

app/endpoints/campaign.py
```diff
@@ -716,14 +769,87 @@ async def create_campaigns_logo(
...
+    status_code=201,
+    tags=[Tags.campaign],
+)
+async def edit_voters(
+    voters_id: str,
+    voters_update: schemas_campaign.Voters,
+    user: models_core.CoreUser = Depends(is_user_a_member_of(GroupType.CAA)),
+    db: AsyncSession = Depends(get_db),
+):
+    await cruds_campaign.update_voters(id=voters_id, voters_update=voters_update, db=db)
```
armanddidierjean, line 838: Is there really a need for this?

armanddidierjean (changes requested):

app/models/models_campaign.py
```diff
@@ -69,3 +69,10 @@ class Status(Base):
 
     id: Mapped[str] = mapped_column(String, nullable=False, primary_key=True)
     status: Mapped[StatusType] = mapped_column(Enum(StatusType), nullable=False)
+
+
+class Voters(Base):
+    __tablename__ = "campaign_voters"
+
+    id: Mapped[str] = mapped_column(String, nullable=False, primary_key=True)
```
armanddidierjean, line 77: A group should not be present more than one time, we could remove the id

Petitoto (approved):

app/endpoints/campaign.py
```diff
@@ -320,6 +345,66 @@ async def update_list(
...
+
+
+@router.post(
+    "/campaign/voters",
+    response_model=list[schemas_campaign.VoterGroup],
+    status_code=201,
+    tags=[Tags.campaign],
+)
+async def add_voters(
+    voters: list[schemas_campaign.VoterGroup],
```
armanddidierjean, line 372: I think it may be easier to pass a list of id directly.

For example:
```suggestion
    allowed_group_ids: list[str],
```

app/endpoints/campaign.py
```diff
@@ -492,15 +577,23 @@ async def reset_vote(
...
     Add a vote for a given campaign list.
 
     An user can only vote for one list per section.
 
-    **The user must be a member of the group AE to use this endpoint**
+    **The user must be a voter to use this endpoint**
     """
+    voters = await cruds_campaign.get_voters(db)
+    voters_groups = [voter.group_id for voter in voters]
+    if not voters_groups or not is_user_member_of_an_allowed_group(user, voters_groups):
```
armanddidierjean, line 591: We either want to allow everyone to vote in `get_lists` or nobody here, when `voters_groups` is empty
```suggestion
    if not is_user_member_of_an_allowed_group(user, voters_groups):
```

app/models/models_campaign.py
```diff
@@ -45,6 +45,12 @@ def as_dict(self):
         return {c.name: getattr(self, c.name) for c in self.__table__.columns}
 
 
+class VoterGroups(Base):
```
armanddidierjean, line 48: I'm not sure `VoterGroups` is explicit enough for someone who does not know the module, to be honest.

armanddidierjean: I don't know why we put plural for every model in this file, but we usually use singular names.
```suggestion
class VoterGroup(Base):
```

app/models/models_campaign.py
```diff
@@ -45,6 +45,12 @@ def as_dict(self):
...
+    __tablename__ = "campaign_votergroups"
```
armanddidierjean, line 49:
```suggestion
    __tablename__ = "campaign_voter_groups"
```

armanddidierjean (changes requested):

app/cruds/cruds_campaign.py
```diff
@@ -36,6 +36,31 @@ async def get_status(
     return status[0].status
 
 
+async def get_voters(db: AsyncSession) -> Sequence[models_campaign.VoterGroups]:
+    result = await db.execute(select(models_campaign.VoterGroups))
+    return result.scalars().all()
+
+
+async def add_voters(
```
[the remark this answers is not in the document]
Petitoto: Related to the question of renaming voters to something else

app/endpoints/campaign.py
Petitoto: We will pass a single VoterGroup

app/models/models_campaign.py
Petitoto: > I'm not sure `VoterGroups` is explicit enough for someone who does not know the module, to be honest.

I'll add comments. No better name found for now

Petitoto: > I don't know why we put plural for every model in this file, but we usually use singular names.

To avoid useless migrations, i would prefer staying with plural for every model in this file

PR #270, opened 2023-10-18
Use bcrypt instead of passlib

Jake-Ransom:
### Description

Use the bcrypt library for password hashing instead of passlib that used a deprecated dependency (crypt).

### Checklist

- [x] All tests passing

### DIFF
--- a/app/core/security.py
+++ b/app/core/security.py
@@ -1,9 +1,9 @@
 import secrets
 from datetime import datetime, timedelta
 
+import bcrypt
 from fastapi.security import OAuth2AuthorizationCodeBearer
 from jose import jwt
-from passlib.context import CryptContext
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from app.core.config import Settings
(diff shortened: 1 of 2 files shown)


app/core/security.py
```diff
@@ -1,17 +1,16 @@
...
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from app.core.config import Settings
 from app.cruds import cruds_users
 from app.models import models_core
 from app.schemas import schemas_auth
 
-pwd_context = CryptContext(schemes=["bcrypt"], deprecated="auto", bcrypt__rounds=13)
 """
 In order to salt and hash password, we use a *passlib* [CryptContext](https://passlib.readthedocs.io/en/stable/narr/quickstart.html) object.
```
armanddidierjean, line 15: We should just update this comment as we don't use Passlib anymore

armanddidierjean (approved):

Petitoto (approved):

PR #280, opened 2023-11-04
 Improving details of booking showed when tapped on calendar

julien4215:
Required for PR https://github.com/aeecleclair/Titan/pull/185


app/endpoints/booking.py
```diff
@@ -184,7 +184,7 @@ async def get_bookings_for_manager(
 
 @router.get(
     "/booking/bookings/confirmed",
-    response_model=list[schemas_booking.BookingReturn],
+    response_model=list[schemas_booking.BookingReturnApplicant],
```
armanddidierjean, line 187: I'm not sure if we want to return the applicant Phone number and email address

armanddidierjean (changes requested):

app/endpoints/booking.py
julien4215: I didn't think about that, I'll create a new endpoint for managers

julien4215: Yes I didn't pay attention to that.
I actually don't need to make changes to the API.

armanddidierjean (approved):

Rotheem (changes requested):
The migration's file needs to be entirely changed

Rotheem (approved):

Issue #285, opened 2023-11-06
Visibility: register api routes based on ModuleList


armanddidierjean:
### Subject of the issue

Currently, we need to register api routers in one file [/app/main/app/api.py](https://github.com/aeecleclair/Hyperion/blob/main/app/api.py) and register the module visibility in [/app/utils/types/module_list.py](https://github.com/aeecleclair/Hyperion/blob/main/app/utils/types/module_list.py)

PR #289, opened 2023-11-13
Module structure

Pokegali:
### Description

This PR moves all module-related files to `/app/modules`, and each module has its own directory.
It also makes modules declare their own module object.

Note: it is based on pydantic v2 (https://github.com/aeecleclair/Hyperion/pull/281).


app/api.py
```diff
@@ -1,39 +1,39 @@
...
+api_router.include_router(endpoints_notification.router)
+api_router.include_router(endpoints_users.router)
+
+for endpoints_file in glob.glob("./app/modules/*/endpoints_*.py"):
+    endpoint_module = importlib.import_module(endpoints_file[2:-3].replace("/", "."))
+    if hasattr(endpoint_module, "module"):
+        modules = endpoint_module.module
+        if not isinstance(modules, Iterable):
+            modules = [modules]
+        module_list.extend(modules)
```
armanddidierjean, line 33: I'm not sure if this is really a good idea to accept both an Iterable and a single Module

Pokegali:
I've deleted the declaration of the `home` module in the calendar, because that's not how it should be done

app/api.py
Pokegali: Fixed, now we can only declare one module per file

app/core/module.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def __init__(
+        self,
+        root: str,
+        default_allowed_groups_ids: list[GroupType] | None = None,
+        router: APIRouter | None = None,
+    ):
+        self.root = root
+        self.default_allowed_groups_ids = default_allowed_groups_ids or []
+        self.router = router or APIRouter()
```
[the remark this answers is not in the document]
Pokegali: Fixed, the allowed groups are no longer optional, and the docstring has been added

armanddidierjean:
Should fix #285

app/app.py
```diff
@@ -188,12 +188,13 @@ def initialize_module_visibility(engine: Engine) -> None:
...
-        for module in ModuleList:
-            for default_group_id in module.value.default_allowed_groups_ids:
+        for module in module_list:
+            for default_group_id in module.default_allowed_groups_ids:
                 module_visibility = models_core.ModuleVisibility(
-                    root=module.value.root, allowed_group_id=default_group_id.value
+                    root=module.root, allowed_group_id=default_group_id.value
                 )
                 try:
+                    print(module.root, default_group_id)
```
guillaumetavernier, line 197: Should this `print` remain here ?

Pokegali: Oops

guillaumetavernier (approved):

maximeroucher (approved):

app/core/auth/endpoints_auth.py
```diff
@@ -21,29 +21,29 @@
...
 )
-from app.models import models_auth, models_core
-from app.schemas import schemas_auth
 from app.utils.auth.providers import BaseAuthClient
 from app.utils.tools import is_user_member_of_an_allowed_group
 from app.utils.types.scopes_type import ScopeType
-from app.utils.types.tags import Tags
 
 router = APIRouter()
+tag = "Auth"
```
AtomHare, line 46: We may want to use `router = APIRouter(tags=["Auth"])` instead (see [here](https://fastapi.tiangolo.com/tutorial/bigger-applications/?h=tags#add-some-custom-tags-responses-and-dependencies)) and remove it from each request (this goes for every module)

app/core/module.py
AtomHare, line 1: Like my other comment, do we want to support tags natively like `APIRouter` ? When we instantiate the module the tag could be set thus removing the need to use `tags=[tag]` on each endpoint

AtomHare (commented):
See my comments otherwise LGTM :+1: 
Nice PR !

app/core/module.py
Pokegali: That's a good idea, let me try that

AtomHare (approved):

Pokegali:
It looks like we all missed it, but we should also maybe refactor the notifications, so that each module declares its own topic, instead of them being declared in app/core/notification/notification_types

app/api.py
```diff
@@ -1,39 +1,38 @@
...
+api_router.include_router(endpoints_auth.router)
+api_router.include_router(endpoints_core.router)
+api_router.include_router(endpoints_groups.router)
+api_router.include_router(endpoints_notification.router)
+api_router.include_router(endpoints_users.router)
+
+for endpoints_file in Path(".").glob("app/modules/*/endpoints_*.py"):
+    endpoint_module = importlib.import_module(
+        ".".join(endpoints_file.with_suffix("").parts)
+    )
```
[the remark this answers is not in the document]
Pokegali: The reason is that in the migrations case I just want to import the module, but in this case I also want the module object

armanddidierjean (approved):

PR #312, opened 2023-12-19
Run Alembic migrations on Hyperion startup

Rotheem:
### Description

Before this pull request we only created missing tables on Hyperion startup. The developer had to run migrations files mannually when a table was modified.

With this PR, Hyperion:
 - create an up to date database and stamp it as "head" when the database was never initialized
 - upgrade the database to "head" by asking alembic to run migration files


armanddidierjean:
We may want to experiment with https://alembic.sqlalchemy.org/en/latest/cookbook.html#building-an-up-to-date-database-from-scratch

Petitoto (approved):

armanddidierjean (approved):

PR #320, opened 2024-01-08
Delete revoked refresh tokens on successful authentication

Jake-Ransom:


app/cruds/cruds_auth.py
```diff
@@ -108,3 +108,18 @@ async def revoke_refresh_token_by_client_and_user_id(
...
+    """Delete expired refresh tokens of a specific user"""
+
+    await db.execute(
+        delete(models_auth.RefreshToken).where(
+            models_auth.RefreshToken.expire_on
+            < datetime.now(timezone(settings.TIMEZONE))
+        )
+    )
+    await db.commit()
+    return None
```
armanddidierjean, line 125: This function delete all expired tokens, instead of tokens from a specific user, but I kind of like this behaviour.

PR #336, opened 2024-02-12
Fix migrations & initializations when using multiple workers

Petitoto:
### Description

This PR fixes migrations when working with multiple workers.

It allows to run migrations and initializations in a synchronous way before lifespan. Thus, gunicorn can preload the app to run them before creating workers.

### Checklist

- [x] All tests passing


requirements.txt
```diff
@@ -12,5 +12,6 @@ unidecode==1.3.7
 uvicorn[standard]==0.23.2
 redis==5.0.1
 icalendar == 5.0.11
+psycopg2==2.9.9
```
armanddidierjean, line 15: Could you add a comment explaining what the dependency is used for?

app/app.py
```diff
@@ -261,6 +244,26 @@ async def startup(app: FastAPI):
...
+    initialize_groups(sync_engine)
+    initialize_module_visibility(sync_engine)
+
+    # Initialize Redis
+    if (
+        app.dependency_overrides.get(get_redis_client, get_redis_client)(
+            settings=settings
+        )
+        is False
+    ):
```
armanddidierjean, line 264:
```suggestion
    if (
        not app.dependency_overrides.get(get_redis_client, get_redis_client)(
            settings=settings
        )
    ):
```

app/app.py
```diff
@@ -261,6 +244,26 @@ async def startup(app: FastAPI):
         allow_headers=["*"],
     )
 
+    # Initialize database connection
+    app.dependency_overrides.get(get_db_engine, get_db_engine)(settings=settings)
```
armanddidierjean, line 248: Is this useful? We don't use the returned engine.

armanddidierjean (changes requested):

app/app.py
Petitoto: It creates the Async Engine, which is stored as a global variable and then used by all workers

PR #344, opened 2024-02-23
External account

BLEMENT33:
### Description

I created code to be able to delete accounts that do not have a Central address.
To do that, I :
* Added a Boolean enabled attribute to the CoreUser
* Added a CRUD to deactivate a CoreUser
* Have prevented disabled CoreUsers from connecting

I still have to exclude deactivated CoreUsers from the search

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -13,7 +13,9 @@
     "[python]": {
         "editor.defaultFormatter": "ms-python.black-formatter"
     },
-    "python.testing.pytestArgs": ["tests"],
+    "python.testing.pytestArgs": [
+        "."
+    ],
     "python.testing.unittestEnabled": false,
     "python.testing.pytestEnabled": true,
     "black-formatter.path": [
--- a/app/endpoints/auth.py
+++ b/app/endpoints/auth.py
@@ -81,6 +81,11 @@ async def login_for_access_token(
             detail="Incorrect login or password",
             headers={"WWW-Authenticate": "Bearer"},
         )
+    if not user.enabled:
+        raise HTTPException(
+            status_code=status.HTTP_401_UNAUTHORIZED,
+            detail="Disabled account. Contact eclair@myecl.fr for more informations.",
+        )
     # We put the user id in the subject field of the token.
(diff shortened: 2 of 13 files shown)


.vscode/settings.json
```diff
@@ -13,7 +13,9 @@
     "[python]": {
         "editor.defaultFormatter": "ms-python.black-formatter"
     },
-    "python.testing.pytestArgs": ["tests"],
+    "python.testing.pytestArgs": [
+        "."
+    ],
     "python.testing.unittestEnabled": false,
```
armanddidierjean, line 19: Should this be changed?

app/endpoints/auth.py
```diff
@@ -81,6 +81,11 @@ async def login_for_access_token(
             detail="Incorrect login or password",
             headers={"WWW-Authenticate": "Bearer"},
         )
+    if not user.enabled:
+        raise HTTPException(
+            status_code=status.HTTP_401_UNAUTHORIZED,
+            detail="Disabled account. Contact eclair@myecl.fr for more informations.",
```
armanddidierjean, line 87: Do we want to hard code an email in the API response?

app/endpoints/external_account.py
```diff
@@ -0,0 +1,22 @@
...
+from app.utils.types.tags import Tags
+
+router = APIRouter()
+
+
+@router.get(
+    "/external/",
+    status_code=200,
+    tags=[Tags.external_account],
+)
```
armanddidierjean, line 17: I think we could put the endpoint in `users.py`

app/models/models_core.py
```diff
@@ -35,6 +35,7 @@ class CoreUser(Base):
     phone: Mapped[str | None] = mapped_column(String)
     floor: Mapped[FloorsType] = mapped_column(Enum(FloorsType), nullable=False)
     created_on: Mapped[datetime | None] = mapped_column(DateTime(timezone=True))
+    enabled: Mapped[bool] = mapped_column(Boolean, nullable=False)
```
armanddidierjean, line 38: What do you think of `disabled`?

armanddidierjean (changes requested):
Nice module!

PR #352, opened 2024-03-01
Add returned_date field in loan module

julien4215:
This change is required for the PR https://github.com/aeecleclair/Titan/pull/310.

It should be merged after #360

### DIFF
--- a/app/modules/loan/cruds_loan.py
+++ b/app/modules/loan/cruds_loan.py
@@ -1,14 +1,10 @@
 from collections.abc import Sequence
-from datetime import UTC, datetime
 
-from fastapi import Depends
 from sqlalchemy import delete, func, select, update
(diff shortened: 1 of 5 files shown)


app/modules/loan/cruds_loan.py
```diff
@@ -190,13 +190,11 @@ async def update_loan(
 
 async def update_loan_returned_status(
     loan_id: str,
-    returned: bool,
     db: AsyncSession,
+    values: dict[str, object],
```
guillaumetavernier, line 194: Maybe it would be cleaner to pass `returned` and `returned_date` independently as function arguments

guillaumetavernier: fixed

guillaumetavernier (approved):

armanddidierjean (approved):

PR #353, opened 2024-03-02
Tests notifications and send notifications for loan and booking

armanddidierjean:
### Description
The PR add tests for notifications. Due to the usage of Firebase messaging to send notifications, tests can not check notifications can be sent. Instead, they make sure our endpoints to register/unregister a device or subscribe to topics work.

Notifications were added:
 - when a loan is created, to notify the borrower
 - when a booking is accepted or refused to notify the applicant

To prevent anyone from subscribing to `bookingadmin` topic and then be notified of all booking requests, the notification is now send to all members of the room manager group. We way want to send the notification to just a few persons instead of the whole group.

### DIFF
--- a/app/core/notification/endpoints_notification.py
+++ b/app/core/notification/endpoints_notification.py
@@ -254,12 +254,12 @@ async def get_topic(
 
 
 @router.get(
-    "/notification/topics/{topic_str}",
+    "/notification/topics/{topic}",
     status_code=200,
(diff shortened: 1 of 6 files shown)


Foukki:
Sorry I stole everything, pls see : https://github.com/aeecleclair/Hyperion/pull/412.

armanddidierjean:
Superseded by #412

PR #359, opened 2024-03-02
Refactor module visibility to add visibility for new modules

armanddidierjean:
### Description

With this PR, modules visibility are stored with a column `visible: bool` in db.

This allows at startup to add default group ids for new modules, instead of only adding them during the database first initialization

**Breaking change**: this PR will delete existing Visibilities

### DIFF
--- a/app/app.py
+++ b/app/app.py
@@ -186,22 +186,20 @@ def initialize_module_visibility(engine: Engine) -> None:
 
     with Session(engine) as db:
         for module in module_list:
-            existing_module_visibility = (
-                initialization.get_all_module_visibility_by_root_sync(
-                    root=module.root,
-                    db=db,
-                )
+            module_awareness = initialization.get_module_awareness_by_root(
+                root=module.root,
+                db=db,
             )
-            if len(existing_module_visibility):
-                # There are already module visibilities for this module
-                # we don't want to override them with default group ids
+            if module_awareness is None:
+                # The module is known by Hyperion
+                # This mean visibilities for the module were already added
+                # We don't want to override custom visibility with default group ids
                 continue
             # We add the module visibility for the default groups
             for default_group_id in module.default_allowed_groups_ids:
                 module_visibility = models_core.ModuleVisibility(
                     root=module.root,
                     allowed_group_id=default_group_id.value,
-                    visible=True,
                 )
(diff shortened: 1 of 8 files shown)


Petitoto (approved):

app/core/models_core.py
```diff
@@ -107,6 +107,7 @@ class ModuleVisibility(Base):
 
     root: Mapped[str] = mapped_column(String, primary_key=True)
     allowed_group_id: Mapped[str] = mapped_column(String, primary_key=True)
+    visible: Mapped[bool] = mapped_column(Integer, nullable=False, default=True)
```
Rotheem, line 110: This should be a boolean and not an Integer

Rotheem (changes requested):

app/core/endpoints_core.py
```diff
@@ -155,17 +155,16 @@ async def get_module_visibility(
...
     """
-    Get all existing module_visibility.
+    Get group ids allowed to see a module for all existing modules in Hyperion.
 
     **This endpoint is only usable by administrators**
     """
-    from app.api import (  # We need to do it here, because the list is not initialized at startup
-        module_list,
-    )
+    # We need to import the module_list here, because the it is not initialized at startup
```
julien4215, line 162:
```suggestion
    # We need to import the module_list here, because it is not initialized at startup
```

app/app.py
```diff
@@ -179,20 +179,17 @@ def initialize_module_visibility(engine: Engine) -> None:
...
-
-        hyperion_error_logger.info(
-            "Startup: Modules visibility settings are empty, initializing them",
-        )
         for module in module_list:
+            module_awareness = initialization.get_module_awareness_by_root(
+                root=module.root,
+                db=db,
+            )
+            if module_awareness is None:
```
julien4215, line 187:
```suggestion
            if module_awareness is not None:
```

julien4215 (changes requested):

julien4215:
Instead of deleting the ModuleVisibility table, the migration file could add all modules in the table ModuleAwarness because all these modules already use visibily

armanddidierjean:
Completed in https://github.com/aeecleclair/Hyperion/pull/612

PR #360, opened 2024-03-02
Fix datetime stored in database

julien4215:
Currently Titan gives to Hyperion local datetime that we are storing in the database as timestamp with timezone. The problem is that they are never converted to utc so we have wrong values in the database. 

We should check if a timezone is set in the `postgresql.conf` and set it to Etc/UTC if it is not the case.

Tracking issue: https://github.com/aeecleclair/Titan/issues/114


guillaumetavernier (commented):
We can still find reference to `TIMEZONE` in docker compose files and in config.py. Did you check those ?

julien4215:
> We can still find reference to `TIMEZONE` in docker compose files and in config.py. Did you check those ?

Yes I will add again the timezone parameter in .env as we still need it to configure database timezone.

julien4215:
I am converting this PR to draft until we decide if we keep supporting both postgresql and sqlite.

app/utils/types/datetime.py
```diff
@@ -0,0 +1,21 @@
+import datetime
+
+from sqlalchemy import DateTime
+from sqlalchemy.types import TypeDecorator
+
+
+class TZDateTime(TypeDecorator):
```
armanddidierjean, line 7: Could you add comments explaining when are called these two functions and where is the class used?

armanddidierjean (approved):

guillaumetavernier (approved):

PR #366, opened 2024-03-15
Use integer cents in AMAP / Tombola

BLEMENT33:
### Description

I change all the amounts in cents (integer) to avoid rounding issues.

### DIFF
--- a/app/modules/raffle/endpoints_raffle.py
+++ b/app/modules/raffle/endpoints_raffle.py
@@ -189,9 +189,9 @@ async def get_raffle_stats(
     tickets = await cruds_raffle.get_tickets_by_raffleid(db=db, raffle_id=raffle_id)
 
     tickets_sold = len(tickets)
-    amount_raised = sum(
+    amount_raised = int(sum(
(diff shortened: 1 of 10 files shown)


app/modules/raffle/endpoints_raffle.py
```diff
@@ -189,9 +189,9 @@ async def get_raffle_stats(
     tickets = await cruds_raffle.get_tickets_by_raffleid(db=db, raffle_id=raffle_id)
 
     tickets_sold = len(tickets)
-    amount_raised = sum(
+    amount_raised = int(sum(
         [ticket.pack_ticket.price / ticket.pack_ticket.pack_size for ticket in tickets]
-    )
+    ))
```
guillaumetavernier, line 194: The goal of this PR is to avoid imprecision errors. Maybe there is a way to avoid dividing values before converting to int

guillaumetavernier:
Nice one ! You will need to look into migration files to convert float euro database values to new int cent values

armanddidierjean:
Superceded by https://github.com/aeecleclair/Hyperion/pull/686 for AMAP module

PR #368, opened 2024-03-16
PH v1

BLEMENT33:
### Description

I sarted a PH module for the journalism association of ECL.


armanddidierjean (commented):
Nice module!

Could you change endpoints to use a lowercase `ph`? Maybe "newspaper" or "paper" would be better instead of "journal"

app/modules/ph/endpoints_ph.py
```diff
@@ -0,0 +1,112 @@
...
+
+@module.router.get(
+    "/ph/",
+    response_model=list[schemas_ph.PaperComplete],
+    status_code=200,
+)
+async def get_papers(
+    db: AsyncSession = Depends(get_db),
+    user: models_core.CoreUser = Depends(is_user_a_member),
+):
```
armanddidierjean, line 50: Could you add a docstring explaining the results are sorted by date?

app/modules/ph/endpoints_ph.py
```diff
@@ -0,0 +1,112 @@
...
+@module.router.get(
+    "/ph/{paper_id}/pdf",
+    response_class=FileResponse,
+    status_code=200,
+)
+async def get_paper_pdf(
+    paper_id: str,
+    user: models_core.CoreUser = Depends(is_user_a_member),
+):
+    return get_file_from_data(
```
armanddidierjean, line 35: You must make sure that paper_id is a valid ph id before calling this function. Currently someone could use this endpoint to access files from other modules

tests/test_PH.py
```diff
@@ -0,0 +1,78 @@
...
+from app.modules.ph import models_ph
+from tests.commons import (
+    add_object_to_db,
+    client,
+    create_api_access_token,
+    create_user_with_groups,
+    event_loop,  # noqa
+)
+
+
```
armanddidierjean, line 17: I think we need to define global variables here, but I'm not sure

```suggestion
ph_user_ph: models_core.CoreUser
ph_user_simple: models_core.CoreUser
token_simple: str
```

armanddidierjean (changes requested):

app/modules/ph/endpoints_ph.py
BLEMENT33: In the crud or in the endpoint ?

armanddidierjean: I would suggest in both the crud and the endpoint

tests/test_PH.py
```diff
@@ -0,0 +1,78 @@
...
+def test_create_paper_pdf():
+    with Path("assets/pdf/default_PDF.pdf").open("rb") as pdf:
+        response = client.post(
+            f"/ph/{paper.id}/pdf",
+            files={"pdf": ("test_paper.pdf", pdf, "application/pdf")},
+            headers={"Authorization": f"Bearer {token_ph}"},
+        )
+
+    assert response.status_code == 201
```
armanddidierjean, line 78:
```suggestion
    assert response.status_code == 201
    assert Path(f"data/ph/{paper.id}.pdf").is_file()
```

armanddidierjean (changes requested):

app/modules/ph/endpoints_ph.py
```diff
@@ -0,0 +1,163 @@
...
+    paper_complete = schemas_ph.PaperComplete(
+        id=str(uuid.uuid4()),
+        **paper.model_dump(),
+    )
+    try:
+        paper_db = models_ph.Paper(
+            id=paper_complete.id,
+            name=paper_complete.name,
+            release_date=paper_complete.release_date,
+        )
```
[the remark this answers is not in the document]
BLEMENT33: I wanted to use schemas_ph.paperComplete because it's not used anywhere else

app/modules/ph/endpoints_ph.py
```diff
@@ -0,0 +1,231 @@
...
+        directory="ph/pdf",
+        filename=str(paper_id),
+        request_id=request_id,
+        max_file_size=10 * 1024 * 1024,
+        accepted_content_types=["application/pdf"],
+    )
+    Path("data/ph/cover/").mkdir(parents=True, exist_ok=True)
+    paper_pdf = fitz.open(f"data/ph/pdf/{paper_id}.pdf")
+    page = paper_pdf.load_page(0)
+    pix = page.get_pixmap(matrix=fitz.Matrix(300 / 72, 300 / 72))
```
armanddidierjean, line 152: Could you add a comment explaining how does the matrix multiplication works, and why the value `300/72` was chosen?

armanddidierjean: Are you sure it's useful to increase the image size by 4? It seams pretty big for a cover

app/utils/tools.py
armanddidierjean, line 1: Is it intended to remove the new modifications about uuid regex and return types?

app/utils/tools.py
```diff
@@ -190,28 +181,36 @@ def get_file_from_data(
...
 
+def delete_file_from_data(
+    directory: str,
+    filename: str,
+    extension: str,
+):
+    """
+    If there is a file with the provided filename in the data folder, delete it. The file extension must be precised.
+    > "data/{directory}/{filename}.{extension}"
+    """
```
armanddidierjean, line 206: Could you add the UUID regex verification and the comment explaining to never thrust the user input?

armanddidierjean (changes requested):

app/utils/tools.py
BLEMENT33: I don't know why but it doesn't work with `"data/{directory}/{filename}.*"` but it works like this so ...

app/utils/tools.py
```diff
@@ -190,28 +181,36 @@ def get_file_from_data(
...
     for filePath in Path().glob(f"data/{directory}/{filename}.*"):
         return FileResponse(filePath)
 
     return FileResponse(default_asset)
 
 
+def delete_file_from_data(
+    directory: str,
+    filename: str,
+    extension: str,
```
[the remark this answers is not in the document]
BLEMENT33: I don't know why but it doesn't work with `"data/{directory}/{filename}.*"`` but it works like this so ...

app/modules/ph/endpoints_ph.py
```diff
@@ -0,0 +1,231 @@
...
+    cover = Image.frombytes("RGB", [pix.width, pix.height], pix.samples)
```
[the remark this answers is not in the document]
BLEMENT33: Like this ? : 
```python
    with fitz.open(f"data/ph/pdf/{paper_id}.pdf") as paper_pdf:
        page = paper_pdf.load_page(0)
        cover = page.get_pixmap()
        await save_file_as_data(
            image=cover,
            directory="ph/cover",
            filename=str(paper_id),
            request_id=request_id,
            max_file_size=10 * 1024 * 1024,  # 10 MB
            accepted_content_types=["image/jpeg"],
        )
```

app/modules/ph/endpoints_ph.py
```diff
@@ -0,0 +1,231 @@
...
+    await save_file_as_data(
+        image=pdf,
+        directory="ph/pdf",
+        filename=str(paper_id),
+        request_id=request_id,
+        max_file_size=10 * 1024 * 1024,
+        accepted_content_types=["application/pdf"],
+    )
+    Path("data/ph/cover/").mkdir(parents=True, exist_ok=True)
+    paper_pdf = fitz.open(f"data/ph/pdf/{paper_id}.pdf")
```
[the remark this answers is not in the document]
BLEMENT33: This is better ? : 
```python
with fitz.open(f"data/ph/pdf/{paper_id}.pdf") as paper_pdf:
    page = paper_pdf.load_page(0)
    cover = page.get_pixmap()
    await save_file_as_data(
        image=cover,
        directory="ph/cover",
        filename=str(paper_id),
        request_id=request_id,
        max_file_size=10 * 1024 * 1024,  # 10 MB
        accepted_content_types=["image/jpeg"],
    )
```

app/modules/ph/endpoints_ph.py
BLEMENT33: I just took a basic example on the internet without understand it for the tests but now, i modified it and it's better.

app/utils/tools.py
BLEMENT33: No, I fixed it

app/modules/ph/endpoints_ph.py
```diff
@@ -0,0 +1,263 @@
...
+                title=f"📗 PH - {paper_db.name}",
+                content="A new paper has been released! 🎉",
+                delivery_datetime=datetime.fromisoformat(
+                    paper_db.release_date.isoformat(),
+                ),
+                # The notification will expire in 10 days
+                expire_on=now + timedelta(days=10),
+            )
+            await notification_tool.send_notification_to_topic(
+                custom_topic=CustomTopic(topic=Topic.advert),
```
guillaumetavernier, line 137: I think that there is the wrong topic here

guillaumetavernier (changes requested):
Not much to say ! Nice work !

app/modules/ph/cruds_ph.py
```diff
@@ -0,0 +1,69 @@
...
+    db: AsyncSession,
+    end_date: date = date(2099, 12, 31),
+) -> Sequence[models_ph.Paper]:
+    """Return papers from the latest to the oldest"""
+    result = await db.execute(
+        select(models_ph.Paper)
+        .where(models_ph.Paper.release_date <= end_date)
+        .order_by(models_ph.Paper.release_date.desc()),
+    )
+    return result.scalars().all()
```
armanddidierjean, line 21: I'm not completely fond of this magic value `31/12/2099`.
Maybe you could do:
```suggestion
    end_date: date | None = None,
) -> Sequence[models_ph.Paper]:
    """Return papers from the latest to the oldest"""
    result = await db.execute(
        select(models_ph.Paper)
        .where(models_ph.Paper.release_date <= (end_date or date.max))
        .order_by(models_ph.Paper.release_date.desc()),
    )
    return result.scalars().all()
```

app/modules/ph/endpoints_ph.py
```diff
@@ -0,0 +1,263 @@
...
+        )
+
+        now = datetime.now(UTC)
+
+        if paper_db.release_date >= now.date() - timedelta(days=30):
+            message = Message(
+                context=f"ph-{paper_db.id}",
+                is_visible=True,
+                title=f"📗 PH - {paper_db.name}",
+                content="A new paper has been released! 🎉",
```
armanddidierjean, line 129: We currently don't support localization for the notification content. Maybe you could write the sentence in French.

app/modules/ph/endpoints_ph.py
```diff
@@ -0,0 +1,263 @@
...
+    try:
+        paper_db = models_ph.Paper(
+            id=paper_complete.id,
+            name=paper_complete.name,
+            release_date=paper_complete.release_date,
+        )
+
+        now = datetime.now(UTC)
+
+        if paper_db.release_date >= now.date() - timedelta(days=30):
```
armanddidierjean, line 124:
```suggestion
        # We only want to send a notification if the paper was released less than a month ago.
        if paper_db.release_date >= now.date() - timedelta(days=30):
```

app/modules/ph/endpoints_ph.py
```diff
@@ -0,0 +1,263 @@
...
+            message = Message(
+                context=f"ph-{paper_db.id}",
+                is_visible=True,
+                title=f"📗 PH - {paper_db.name}",
+                content="A new paper has been released! 🎉",
+                delivery_datetime=datetime.fromisoformat(
+                    paper_db.release_date.isoformat(),
+                ),
```
armanddidierjean, line 132: Maybe we could send the notification during the day, instead of midnight
```suggestion
                delivery_datetime=datetime.combine(paper_db.release_date, time(hour=8, tzinfo=UTC)),
```

app/modules/ph/endpoints_ph.py
```diff
@@ -0,0 +1,263 @@
...
+            detail="The paper does not exist.",
+        )
+
+    await save_file_as_data(
+        upload_file=pdf,
+        directory="ph/pdf",
+        filename=str(paper_id),
+        request_id=request_id,
+        max_file_size=10 * 1024 * 1024,  # 10 MB
+        accepted_content_types=["application/pdf"],
```
armanddidierjean, line 173:
```suggestion
        accepted_content_types=ContentType.pdf,
```

app/modules/ph/endpoints_ph.py
```diff
@@ -0,0 +1,263 @@
...
+    if Path.exists(Path(f"data/ph/pdf/{paper_id}.pdf")):
+        delete_file_from_data(
+            directory="ph/pdf",
+            filename=str(paper_id),
+        )
+    if Path.exists(Path(f"data/ph/cover/{paper_id}.jpg")):
+        delete_file_from_data(
+            directory="ph/cover",
+            filename=str(paper_id),
+        )
```
armanddidierjean, line 258: I think you can call the method without checking if the file exists, `delete_file_from_data` should do that for you.

app/modules/ph/models_ph.py
```diff
@@ -0,0 +1,14 @@
+from datetime import date
+
+from sqlalchemy import Date, String
+from sqlalchemy.orm import Mapped, mapped_column
+
+from app.types.sqlalchemy import Base
+
+
+class Paper(Base):
+    __tablename__ = "papers"
```
armanddidierjean, line 10: Please prefix the table name with `ph` to prevent collision with other modules
```suggestion
    __tablename__ = "ph_paper"
```

armanddidierjean (changes requested):

app/modules/ph/endpoints_ph.py
BLEMENT33: I did that because you said : "We may want to delete a PH that does not have a PDF but use the default PDF"

app/modules/ph/endpoints_ph.py
BLEMENT33: Are you sure ? Because save_file_as_data need a list of string.
https://github.com/aeecleclair/Hyperion/blob/9b6022c825a0203e9aaf693b61ff08ced00454b2/app/utils/tools.py#L107

armanddidierjean: I think the branch is not up-to-date. This should be an enum: https://github.com/aeecleclair/Hyperion/blob/a8a265759ccc2b2af6cc40dcda125ced8ae8c6e1/app/utils/tools.py#L133
I suggest you to merge main into this branch

Foukki (approved):

armanddidierjean (approved):

guillaumetavernier (approved):

PR #372, opened 2024-03-16
Make Hyperion installation easier

julien4215:
Additional steps may be needed to install psycopg2 if we don't use the psycopg2-binary in the requirements.txt (see the documentation https://www.psycopg.org/docs/install.html).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,6 +15,6 @@ unidecode==1.3.8
 uvicorn[standard]==0.23.2
 redis==5.0.2
 icalendar == 5.0.11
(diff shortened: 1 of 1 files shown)


AtomHare:
According to their documentation it looks a bit of an issue on our workflow since we use the same requirements.txt when building Hyperion images.

> If you are the maintainer of a published package depending on psycopg2 you shouldn’t use psycopg2-binary as a module dependency. For production use you are advised to use the source distribution.

julien4215:
Yes it is not a good idea as the requirement.txt is used in production. Maybe we should try to completely separate the two requirements files. It would also be nice to use psycopg 3 instead of 2.

PR #375, opened 2024-03-16
Feat : Add base to allow factory setup

Rotheem:
### Description

### Checklist

- [ ] All tests passing


app/app.py
```diff
@@ -206,6 +207,63 @@ def initialize_module_visibility(engine: Engine) -> None:
...
+        )
+        if hasattr(factory_module, "factory"):
+            factory = factory_module.factory
+            factories_list.append(factory)
+        else:
+            hyperion_error_logger.error(
+                f"Module {factories_file} does not declare a module. It won't be enabled.",
+            )
+
+    action = True
```
armanddidierjean, line 233: I think we should find a solution a bit less complicated

armanddidierjean (changes requested):
I think factories object should be classes inherit from the base Factory class

.vscode/settings.json
```diff
@@ -3,7 +3,7 @@
     "editor.tabSize": 4,
     "editor.insertSpaces": true,
     "editor.detectIndentation": false,
-    "python.analysis.typeCheckingMode": "off",
+    "python.analysis.typeCheckingMode": "standard",
```
armanddidierjean, line 6: This may not be the best PR for this change. I think we initially we disabled typeChecking to force the usage of mypy on vscode

app/app.py
```diff
@@ -216,6 +220,70 @@ def initialize_schools(
                     )
 
 
+async def run_factories(db: AsyncSession, settings: Settings) -> None:
+    """Run the factories to create default data in the database"""
+    hyperion_error_logger = logging.getLogger("hyperion.error")
+    if not settings.USE_FACTORIES:
+        hyperion_error_logger.info("Startup: Factories are disabled")
```
armanddidierjean, line 227: As this feature is not expected to be run in production, I think we don't need to have a log each time the prod server starts.

I would suggest logging a message when factories are enabled and run instead

app/core/core_endpoints/endpoints_core.py
```diff
@@ -19,6 +19,8 @@
 from app.types.module import CoreModule
 from app.utils.tools import is_group_id_valid
 
+hyperion_error_logger = logging.getLogger("hyperion.error")
```
armanddidierjean, line 22: Is this used?

app/utils/factory.py
```diff
@@ -0,0 +1,21 @@
...
+    def __init__(
+        self,
+        name: str,
+        depends_on: list[type["Factory"]],
+    ) -> None:
+        self.name = name
+        self.depends_on = depends_on
+
+    @abstractmethod
+    async def should_run(self, db: AsyncSession) -> bool:
```
armanddidierjean, line 16: Could you document this method?

armanddidierjean (changes requested):

.vscode/settings.json
Rotheem: It was not supposed to be committed.
I activated it as a fast way to have type checking for him

requirements-common.txt
```diff
@@ -5,6 +5,7 @@ asyncpg==0.29.0                     # PostgreSQL adapter for asynchronous operat
 bcrypt==4.1.3                       # password hashing
 broadcaster==0.3.1                  # Working with websockets with multiple workers.
 calypsso==1.2.0
+Faker==37.1.0
```
armanddidierjean, line 8:
```suggestion
Faker==37.1.0                       # Fake data generator for factories 
```

app/app.py
```diff
@@ -218,6 +221,61 @@ def initialize_schools(
...
+            factories_list.append(module.factory)
+        else:
+            hyperion_error_logger.warning(
+                f"Module {module.root} does not declare a factory. It won't provide any base data.",
+            )
+
+    # We have to run the factories in a specific order to make sure the dependencies are met
+    # For that reason, we will run the first factory that has no dependencies, after that we remove it from the list of the dependencies from the other factories
+    # And we loop until there are no more factories to run and we use a boolean to avoid infinite loops with circular dependencies
+    action = True
```
armanddidierjean, line 244: I'm not sure what name would be better, but I don't find `action` really explicit
```suggestion
    run_factory = True
```

app/app.py
```diff
@@ -340,6 +398,94 @@ def init_db(
...
+    # We use the parent process to get the workers
+    parent_pid = os.getppid()  # PID du parent (FastAPI master process)
+    parent_process = psutil.Process(parent_pid)
+    workers = [
+        p for p in parent_process.children() if p.status() != psutil.STATUS_ZOMBIE
+    ]
+
+    # There is more than one worker
+    # We use the redis client to check if the factories have already been run and the google api credentials have been set
+    if len(workers) > 1:
```
armanddidierjean, line 416: Do we really gain from checking this? I think we could directly check if a Redis is available

app/app.py
```diff
@@ -340,6 +398,94 @@ def init_db(
...
+        redis_client = app.dependency_overrides.get(
+            get_redis_client,
+            get_redis_client,
+        )(settings=settings)
+        if type(redis_client) is redis.Redis:
```
armanddidierjean, line 421:
```suggestion
        if type(redis_client) is not redis.Redis:
            hyperion_error_logger.info(
                "Redis client not configured while multiple workers are used. Skipping data initialization.",
            )
        else:
```

app/app.py
```diff
@@ -438,6 +558,8 @@ async def lifespan(app: FastAPI) -> AsyncGenerator:
     else:
         hyperion_error_logger.info("Database initialization skipped")
 
+    # Run factories
+
```
armanddidierjean, line 562:
```suggestion
```

app/core/groups/factory_groups.py
```diff
@@ -0,0 +1,60 @@
...
+
+    async def run(self, db: AsyncSession):
+        await self.create_core_groups(db=db)
+        await self.create_core_memberships(db=db)
+
+    async def should_run(self, db: AsyncSession):
+        return len(await cruds_groups.get_groups(db=db)) == len(GroupType)
+
+
+factory = CoreGroupsFactory()
```
armanddidierjean, line 60: Could we init `CoreGroupsFactory` directly in endpoints_groups?

app/types/factory.py
```diff
@@ -0,0 +1,49 @@
...
+        depends_on: list[type["Factory"]],
+    ) -> None:
+        self.name = name
+        self.depends_on = depends_on
+
+    @abstractmethod
+    async def should_run(self, db: AsyncSession) -> bool:
+        """
+        Check if the factory should run.
+        This prevent duplicate runs of the same factory.
```
armanddidierjean, line 19: Could you add a few words on the reason a factory may want to prevent its execution?

app/types/factory.py
```diff
@@ -0,0 +1,49 @@
+from abc import ABC, abstractmethod
+
+from sqlalchemy.ext.asyncio import AsyncSession
+
+
+class Factory(ABC):
+    def __init__(
+        self,
+        name: str,
```
armanddidierjean, line 9: Is the name of the factory really useful? Could we directly use the name of the class?

app/types/factory.py
```diff
@@ -0,0 +1,49 @@
...
+    @abstractmethod
+    async def should_run(self, db: AsyncSession) -> bool:
+        """
+        Check if the factory should run.
+        This prevent duplicate runs of the same factory.
+        """
+
+    @abstractmethod
+    async def run(self, db: AsyncSession) -> None:
+        pass
```
armanddidierjean, line 24: Please add a small docstring

app/types/factory.py
```diff
@@ -0,0 +1,49 @@
...
+
+    async def run(self, db: AsyncSession):
+        pass
+
+    async def should_run(self, db: AsyncSession):
+        return True
+
+
+factory = ModuleFactory()
+"""
```
armanddidierjean, line 49: This is great. Could you maybe put this in the class docstring?

armanddidierjean (commented):
Nice PR! Could you add tests to ensure that starting from an empty db all factories can be run and are compatible?

.env.template
```diff
@@ -74,11 +74,13 @@ POSTGRES_PASSWORD = ""
 POSTGRES_DB = ""
 POSTGRES_TZ = "Etc/UTC"
 DATABASE_DEBUG = False #If True, will print all SQL queries in the console
+USE_FACTORIES = True #If True, will use factories to populate the database with fake data
+FACTORIES_DEMO_USERS_PASSWORD =  #Password used to create demo users
```
armanddidierjean, line 78:
```suggestion
#USE_FACTORIES = True #If True, will use factories to populate the database with fake data
#FACTORIES_DEMO_USERS_PASSWORD =  #Password used to create demo users
```

app/types/websocket.py
```diff
@@ -307,3 +307,56 @@ async def manage_websocket(
                 connection=websocket,
             )
             raise
+
+
+class OfflineWebsocketConnectionManager(WebsocketConnectionManager):
```
armanddidierjean, line 312: I'm not sure why we would need an OfflineWebsocketConnectionManager. We always need a manager which can use either a redis or an inmemory queue.

armanddidierjean (changes requested):

PR #376, opened 2024-03-16
Only allow content-type based on an enum for saving files

armanddidierjean:
### Checklist

- [x] All tests passing

### DIFF
--- a/app/core/users/endpoints_users.py
+++ b/app/core/users/endpoints_users.py
@@ -31,9 +31,9 @@ from app.dependencies import (
     is_user_a_member,
     is_user_a_member_of,
 )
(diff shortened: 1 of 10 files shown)


app/utils/types/content_type.py
```diff
@@ -0,0 +1,12 @@
+from enum import Enum
+
+
+class ContentType(str, Enum):
```
julien4215, line 4: Do we need ContentType to inherit of str ?

julien4215, line 4: Do we need `ContentType` to inherit from `str` ?

julien4215 (approved):

julien4215 (approved):

PR #377, opened 2024-03-16
password requirements

### LINKED ISSUE #330: Set password requirements
We should update our [password validators](https://github.com/aeecleclair/Hyperion/blob/bbeeecc7e7ac5aaed9c1108e75e8e24f6dba6d00/app/utils/validators.py#L12) to unsure new users' password are strong enough. 

This need to be coordinated with Titan to make sur the requirements are the same on both sides

BLEMENT33:
### Description

update our password validators to unsure new users' password are strong enough.
We need to coordinate this with Titan.

Fix: #330


app/utils/validators.py
```diff
@@ -11,8 +11,27 @@ def password_validator(password: str) -> str:
...
+            nb_maj += 1
+        elif i.islower():
+            nb_min += 1
+
+    if (
+        len(password) < 6
+        and nb_number < 1
+        and nb_special < 1
+        and nb_min < 1
+        and nb_maj < 1
```
armanddidierjean, line 30: I think the error should be raised if any of this condition is True

```suggestion
        len(password) < 6
        or nb_number < 1
        or nb_special < 1
        or nb_min < 1
        or nb_maj < 1
```

app/utils/validators.py
```diff
@@ -11,8 +11,27 @@ def password_validator(password: str) -> str:
     https://pydantic-docs.helpmanual.io/usage/validators/#reuse-validators
     """
     # TODO
```
armanddidierjean, line 13: You may remove the TODO

armanddidierjean (changes requested):
Should we add some unit tests?

armanddidierjean (changes requested):

foucblg (approved):

armanddidierjean (approved):

app/utils/validators.py
```diff
@@ -10,9 +10,21 @@ def password_validator(password: str) -> str:
...
+        elif i.isupper():
+            nb_maj += 1
+        elif i.islower():
+            nb_min += 1
+
+    if len(password) < 6 or nb_number < 1 or nb_special < 1 or nb_min < 1 or nb_maj < 1:
+        raise ValueError(
+            "The password must be at least 6 characters long and contain at least one number, one special character, one majuscule and one minuscule.",
+        )
     return password.strip()
```
armanddidierjean, line 28: As remarked by @Rotheem, you should use `strip()` before the validation

armanddidierjean (changes requested):

app/utils/validators.py
Rotheem: And more importantly, use RegEx please

armanddidierjean (changes requested):
We should use the same algorithm  as Calypsso

app/app.py
```diff
@@ -304,6 +307,13 @@ async def lifespan(app: FastAPI) -> AsyncGenerator:
     ):
         hyperion_error_logger.info("Redis client not configured")
 
+    password_dict_files = os.listdir("assets/password_dict")
+    password_dicts: dict[str, list[str]] = {}
+    for password_dict_file in password_dict_files:
+        with Path.open(Path(f"assets/password_dict/{password_dict_file}")) as file:
```
armanddidierjean, line 313:
```suggestion
    for password_dict_file in Path("assets/password_dict").glob("*"):
        with Path.open(password_dict_file) as file:
```

assets/password_dict/commonWords.json
armanddidierjean, line 1: Could we express `assets/password_dict` contains French word lists? Maybe we could create a `fr` subfolder: `assets/password_dict/fr` ?

app/utils/validators.py
```diff
@@ -10,10 +14,20 @@ def password_validator(password: str) -> str:
...
+        raise ValueError(
+            result["feedback"],
+        )
+    if not re.fullmatch(
+        r"(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.*[!@#$€%^&*\(\)_+\-.,.?\":\{\}|<>'/;\[\]]).{6,}",
+        password,
+    ):
+        raise ValueError(
+            "The password must be at least 6 characters long and contain at least one number, one special character, one majuscule and one minuscule.",
+        )
```
armanddidierjean, line 29: We can remove the regex
```suggestion
```

app/app.py
armanddidierjean: If you use a subfolder: `.glob("**/*"):`

app/app.py
```diff
@@ -304,6 +307,13 @@ async def lifespan(app: FastAPI) -> AsyncGenerator:
     ):
         hyperion_error_logger.info("Redis client not configured")
 
+    password_dict_files = os.listdir("assets/password_dict")
```
armanddidierjean, line 310: Could we add a comment explaining why we add these dicts and where they come from?

armanddidierjean (changes requested):

app/utils/validators.py
Rotheem: No we can't as "azdgnuhmokbh" would be a valid password because it has a score of 4 for zxcvbn

PR #382, opened 2024-03-17
Add migration for notifications create_on row

Petitoto:
### Description

Add migration for #371

### Checklist

- [x] All tests passing

### DIFF
--- a/migrations/versions/7-notifications_create_on.py
+++ b/migrations/versions/7-notifications_create_on.py
@@ -22,9 +22,7 @@ def upgrade() -> None:
     op.alter_column(
         "notification_message",
         "expire_on",
-        existing_type=sa.DATE(),
         type_=sa.DateTime(timezone=False),
-        existing_nullable=False,
     )
 
 
@@ -32,7 +30,5 @@ def downgrade() -> None:
     op.alter_column(
         "notification_message",
         "expire_on",
-        existing_type=sa.DateTime(timezone=False),
         type_=sa.DATE(),
-        existing_nullable=False,
     )
--- a/migrations/versions/7-notifications_create_on.py
+++ b/migrations/versions/7-notifications_create_on.py
@@ -1,7 +1,7 @@
 """Notifications 'create_on' migration
 
 Revision ID: e3d06397960d
-Revises: 2fcadbe2f0ad
+Revises: 6afc765adaa2
 Create Date: 2024-03-17 19:19:14.297195
 
 """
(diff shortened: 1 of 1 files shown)


migrations/versions/7-notifications_create_on.py
```diff
@@ -0,0 +1,38 @@
...
+
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "e3d06397960d"
```
julien4215, line 15: The revision should be the one of phonebook
```suggestion
revision: str = "6afc765adaa2"
```

julien4215 (changes requested):
Nice migration file !
Do we need the `existing_type` and `existing_nullable` parameter in `alter_column` ? It seems only required for MySQL according to the documentation.

migrations/versions/7-notifications_create_on.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def upgrade() -> None:
+    op.alter_column(
+        "notification_message",
+        "expire_on",
+        existing_type=sa.DATE(),
+        type_=sa.DateTime(timezone=False),
+        existing_nullable=False,
+    )
```
julien4215, line 28: Nice migration file !
Do we need the existing_type and existing_nullable parameter in alter_column ? It seems only required for MySQL according to the documentation.

julien4215 (approved):

migrations/versions/7-notifications_create_on.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def downgrade() -> None:
+    op.alter_column(
+        "notification_message",
+        "expire_on",
+        existing_type=sa.DateTime(timezone=False),
+        type_=sa.DATE(),
```
julien4215, line 36: Maybe we could use the CamelCase datatype ?

julien4215 (changes requested):
Nice migration file !
Do we need the existing_type and existing_nullable parameter in alter_column ? It seems only required for MySQL according to the documentation.

guillaumetavernier (approved):

migrations/versions/7-notifications_create_on.py
Petitoto: I'm not sure to understand the difference. I commited the change, but did not test.

Foukki (approved):

migrations/versions/7-notifications_create_on.py
julien4215: The uppercase type DATE is the SQL type DATE whereas Date in CamelCase is an agnostic database type meaning that SQLAlechemy will use the the most appropriate SQL type depending on the database. In the documentation they give the exemple of the Boolean that will render BOOLEAN for PostgreSQL, BIT for MySQLand SMALLINT for Oracle.

julien4215 (approved):

Rotheem (approved):

PR #383, opened 2024-03-20
Fix phonebook : kinds enum and types file

Rotheem:
### Description

### Checklist

- [ ] All tests passing

### DIFF
--- a/app/modules/phonebook/types_phonebook.py
+++ b/app/modules/phonebook/types_phonebook.py
@@ -1,7 +1,7 @@
 from enum import Enum
 
 
-class RoleTags(str, Enum):
+class RoleTags(Enum):
     president = "Prez'"
     treso = "Trez'"
     sg = "SG"
@@ -9,7 +9,7 @@ class RoleTags(str, Enum):
     resp_part = "Respo Partenariats"
 
 
-class Kinds(str, Enum):
+class Kinds(Enum):
     comity = "Comité"
     section_ae = "Section AE"
     club_ae = "Club AE"
--- a/app/modules/phonebook/types_phonebook.py
+++ b/app/modules/phonebook/types_phonebook.py
@@ -1,7 +1,7 @@
 from enum import Enum
 
 
-class RoleTags(Enum):
+class RoleTags(str, Enum):
     president = "Prez'"
     treso = "Trez'"
     sg = "SG"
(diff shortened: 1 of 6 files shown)


migrations/versions/6-phonebook.py
```diff
@@ -25,7 +25,19 @@ def upgrade() -> None:
         sa.Column("id", sa.String(), nullable=False),
         sa.Column("name", sa.String(), nullable=False),
         sa.Column("description", sa.String(), nullable=True),
-        sa.Column("kind", sa.String(), nullable=False),
+        sa.Column(
```
guillaumetavernier, line 28: Isn't it possible to use the enum define in type rather than hard code values twice?

tests/test_phonebook.py
```diff
@@ -6,24 +6,20 @@
...
 from app.modules.phonebook import models_phonebook, schemas_phonebook
-from app.modules.phonebook.phonebook_types import Kinds, RoleTags
+from app.modules.phonebook.types_phonebook import Kinds, RoleTags
 from tests.commons import (
     add_object_to_db,
     client,
     create_api_access_token,
     create_user_with_groups,
-    event_loop,  # noqa
+    event_loop,
```
julien4215, line 15: Why did you remove the `# noqa` ?

tests/test_phonebook.py
```diff
@@ -6,24 +6,20 @@
...
 )
 
-association: models_phonebook.Association | None = None
```
julien4215, line 18: Why did you delete these lines ?

julien4215 (changes requested):

app/modules/phonebook/types_phonebook.py
```diff
@@ -1,15 +1,15 @@
 from enum import Enum
 
 
-class RoleTags(Enum):
+class RoleTags(str, Enum):
```
julien4215, line 4: We are not using this Enum as a string so I guess we don't need to inherit from str.

app/modules/phonebook/types_phonebook.py
```diff
@@ -1,15 +1,15 @@
...
     president = "Prez'"
     treso = "Trez'"
     sg = "SG"
     resp_co = "Respo Com'"
     resp_part = "Respo Partenariats"
 
 
-class Kinds(Enum):
+class Kinds(str, Enum):
```
julien4215, line 12: Same here

tests/test_phonebook.py
Rotheem: I think my extensions did weird things

julien4215 (approved):

Foukki (approved):

PR #385, opened 2024-03-25
Send movie db request through Hyperion

Marc-Andrieu:
### Description

Send request to the movie API from Hyperion.

### Checklist

- [x] API key moved here and correctly summoned
- [x] The calls to the movie API work well
- [x] Communicating with Titan works
- [x] Error handling

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -26,7 +26,7 @@ CORS_ORIGINS=["http://localhost"]
 #MATRIX_LOG_AMAP_ROOM_ID = ""
 
 # Token to use the IMDB API
-THE_MOVIE_DB_API = ""
+#THE_MOVIE_DB_API = ""
 
 # SMTP configuration using starttls #
 SMTP_ACTIVE = False
--- a/.env.template
+++ b/.env.template
@@ -25,6 +25,8 @@ CORS_ORIGINS=["http://localhost"]
 #MATRIX_LOG_ERROR_ROOM_ID = ""
 #MATRIX_LOG_AMAP_ROOM_ID = ""
 
+# Token to use the IMDB API
+THE_MOVIE_DB_API = ""
 
 # SMTP configuration using starttls #
 SMTP_ACTIVE = False
--- a/app/modules/cinema/endpoints_cinema.py
+++ b/app/modules/cinema/endpoints_cinema.py
@@ -2,7 +2,7 @@ import logging
 import uuid
 from datetime import timedelta
 
-import requests
+import httpx
 from fastapi import Depends, File, HTTPException, UploadFile
 from fastapi.responses import FileResponse
 from sqlalchemy.ext.asyncio import AsyncSession
@@ -59,14 +59,14 @@ async def get_movie(
         hyperion_error_logger.error("No API key provided for module cinema")
         raise HTTPException(status_code=501, detail="No API key provided")
     try:
-        response = requests.get(
-            url=f"https://api.themoviedb.org/3/movie/{themoviedb_id}",
-            params={
-                "api_key": API_key,
-                "language": "fr-FR",
-            },
-            timeout=5,
-        )
+        async with httpx.AsyncClient() as client:
+            response = await client.get(
+                url=f"https://api.themoviedb.org/3/movie/{themoviedb_id}",
+                params={
+                    "api_key": API_key,
+                    "language": "fr-FR",
+                },
+            )
         if response.status_code != 200:
             hyperion_error_logger.error(
                 f"Code {response.status_code} for IMDb request with movie ID {themoviedb_id}. JSON  response: {response.json()}",
@@ -76,7 +76,7 @@ async def get_movie(
                 detail=f"Movie not found for IMDb movie ID {themoviedb_id} (code {response.status_code})",
             )
         return schemas_cinema.TheMovieDB(**response.json())
-    except requests.RequestException as error:
+    except httpx.RequestError as error:
         hyperion_error_logger.error(error)
         raise HTTPException(status_code=504, detail="Could not reach the IMdB server")
 
--- a/app/modules/cinema/endpoints_cinema.py
+++ b/app/modules/cinema/endpoints_cinema.py
@@ -1,13 +1,13 @@
 import logging
 import uuid
 from datetime import timedelta
-from os import getenv
 from json import loads
-from dotenv import load_dotenv
-from requests import get
+from os import getenv
 
+from dotenv import load_dotenv
 from fastapi import Depends, File, HTTPException, UploadFile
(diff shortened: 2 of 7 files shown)


app/modules/cinema/endpoints_cinema.py
```diff
@@ -28,6 +32,9 @@
 from app.utils.communication.notifications import NotificationTool
 from app.utils.tools import get_file_from_data, save_file_as_data
 
+load_dotenv()
+API_key = getenv("THE_MOVIE_DB_API")
```
armanddidierjean: You can use the class `Settings` and the dependency `get_settings()` to access secrets from the dotenv

app/modules/cinema/endpoints_cinema.py
```diff
@@ -37,6 +44,27 @@
 hyperion_error_logger = logging.getLogger("hyperion.error")
 
 
+@module.router.get(
+    "/cinema/movie/{movie_id}",
+    response_model=schemas_cinema.Movie,
+)
+def get_movie(movie_id):
```
armanddidierjean:
```suggestion
def get_movie(movie_id: str):
```

armanddidierjean: I think this endpoint can be restricted to the group cinema

app/modules/cinema/schemas_cinema.py
```diff
@@ -27,3 +27,11 @@ class CineSessionUpdate(BaseModel):
     overview: str | None = None
     genre: str | None = None
     tagline: str | None = None
+
+class Movie(BaseModel):
+    genres: list[dict] | None = None
```
armanddidierjean, line 33: Would it be possible to make the type a bit more precise than just `dict`?

app/modules/cinema/endpoints_cinema.py
```diff
@@ -37,6 +44,27 @@
...
+    url = f"https://api.themoviedb.org/3/movie/{movie_id}?api_key={API_key}&language=fr-FR"
+    try:
+        r = get(url, timeout=5)
```
armanddidierjean, line 55: You may prefer to use `import requests` and then `requests.get` instead of `get`

armanddidierjean: You can pass query params directly to the `get` method
```suggestion
        response = get(
            url, 
            params: {
                "api_key": API_key,
                "language": "fr-FR",
            },
            timeout=5)
```

app/modules/cinema/endpoints_cinema.py
```diff
@@ -37,6 +44,27 @@
...
+        response = loads(r.content)
+        return schemas_cinema.Movie(
+            genres=response["genres"],
+            overview=response["overview"],
+            poster_path=response["poster_path"],
+            title=response["title"],
+            runtime=response["runtime"],
+            tagline=response["tagline"],
+        )
```
armanddidierjean, line 64:
```suggestion
        return schemas_cinema.Movie(
            **response.json(),
        )
```

app/modules/cinema/endpoints_cinema.py
armanddidierjean: You may also want to explicitly check if the API return an error indicating that the `movie_id` does not exist.
This would be an error caused be Titan that could be returned.

app/modules/cinema/endpoints_cinema.py
```diff
@@ -37,6 +44,27 @@
...
+    except RequestException as error:
+        raise HTTPException(status_code=504, detail=str(error))
```
armanddidierjean, line 66: If another error happens, you may want to log it but return a generic error message to Titan. Sensitive data could indeed be included in the error by themoviedb.
An error 500 could maybe be more adapted.

armanddidierjean (changes requested):

.env.template
```diff
@@ -25,6 +25,8 @@ CORS_ORIGINS=["http://localhost"]
 #MATRIX_LOG_ERROR_ROOM_ID = ""
 #MATRIX_LOG_AMAP_ROOM_ID = ""
 
+# Token to use the IMDB API
+THE_MOVIE_DB_API = ""
```
armanddidierjean, line 29: THE_MOVIE_DB_API is optional and could be commented by default
```suggestion
# Token to use the IMDB API
#THE_MOVIE_DB_API = ""
```

app/modules/cinema/endpoints_cinema.py
```diff
@@ -37,6 +41,31 @@
...
 
+@module.router.get(
+    "/cinema/movie/{movie_id}",
+    response_model=schemas_cinema.Movie,
+)
+def get_movie(
+    movie_id: str,
+    settings: Settings = Depends(get_settings),
+):
+    API_key = settings.THE_MOVIE_DB_API
```
armanddidierjean, line 52: You may want to check if the key is not None, as some Hyperion instances may decide to not include it, and return an error

app/modules/cinema/endpoints_cinema.py
```diff
@@ -37,6 +41,31 @@
 hyperion_error_logger = logging.getLogger("hyperion.error")
 
 
+@module.router.get(
+    "/cinema/movie/{movie_id}",
+    response_model=schemas_cinema.Movie,
+)
+def get_movie(
```
armanddidierjean, line 48: Maybe add a docstring explaining what the method does

app/modules/cinema/endpoints_cinema.py
Marc-Andrieu: [Done](https://github.com/aeecleclair/Hyperion/pull/385/commits/47e184ce13bb6b9a873fc7b536b119d55b4189c6)

Marc-Andrieu:
Guess it's ready for merging? (Merge Titan#338 at the same time)

julien4215 (approved):

app/modules/cinema/endpoints_cinema.py
```diff
@@ -37,6 +40,47 @@
...
+        async with httpx.AsyncClient() as client:
+            response = await client.get(
+                url=f"https://api.themoviedb.org/3/movie/{themoviedb_id}",
+                params={
+                    "api_key": API_key,
+                    "language": "fr-FR",
+                },
+            )
+        if response.status_code != 200:
+            hyperion_error_logger.error(
```
julien4215, line 71: We don't want to log an error if it is due to the client giving a wrong id

julien4215 (changes requested):

julien4215 (approved):

armanddidierjean (approved):

PR #389, opened 2024-03-29
Support external accounts and add stricters group requirements to access endpoints

armanddidierjean:
#### External accounts
With this PR, an `UserCreate` can indicate if Hyperion should accept to create external users. This parameter may be set by third frontend serving a specific purpose, like raid-registering website.

 - if the user has an ECL email address, he will become a Student or Staff, and will be a member (external=False)
 - if he don't have an email address he will still be able to register but won't have access to most modules

#### Dependencies

There are new dependencies:
 - `is_user_a_member` make sure the user is not external
 - `is_user_a_member_of` make sure the user is not external
 - `is_user` don't make any check and should only be used for core functionalities

#### Account Type

We have two account types: 
 - `external` the default group for users that are not members


maximeroucher (approved):

app/core/models_core.py
```diff
@@ -35,9 +35,16 @@ class CoreUser(Base):
     birthday: Mapped[date | None] = mapped_column(Date)
     promo: Mapped[int | None] = mapped_column(Integer)
     phone: Mapped[str | None] = mapped_column(String)
-    floor: Mapped[FloorsType] = mapped_column(Enum(FloorsType), nullable=False)
+    floor: Mapped[FloorsType | None] = mapped_column(Enum(FloorsType))
```
julien4215, line 38: I don't know if we want to make this column nullable since Titan always sends a value

maximeroucher: It was done, because we now allow external account to be created, and they can be created via raid-registering, for which it makes no sense to specify a floor

app/core/groups/groups_type.py
```diff
@@ -16,17 +16,19 @@ class GroupType(str, Enum):
     formerstudent = "ab4c7503-41b3-11ee-8177-089798f1a4a5"
     staff = "703056c4-be9d-475c-aa51-b7fc62a96aaa"
     association = "29751438-103c-42f2-b09b-33fbb20758a7"
+    external = "b1cd979e-ecc1-4bd0-bc2b-4dad2ba8cded"
+    demo = "ae4d1866-e7d9-4d7f-bee7-e0dda24d8dd8"
```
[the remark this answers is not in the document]
Foukki: I guess it is for the AppStore (and PlayStore ?) demo account

Foukki (approved):

PR #390, opened 2024-03-30
Fix OpenAPI specs

Petitoto:
### Description

This PR fixes OpenAPI specs, with the aim to be fully compatible with several API clients like Hoppscotch.

This includes:
- the fix of an invalid endpoint, which should always start with `/` (see https://learn.openapis.org/specification/paths.html#the-endpoints-list)
- the fix of the invalid security scheme_name, which should match `^[a-zA-Z0-9\.\-_]+$` (see https://spec.openapis.org/oas/latest.html#fixed-fields-5)
- automatically set operationId to the path for each route
- add name and version to the OpenAPI specifications
- add `HYPERION_VERSION` and `MINIMAL_TITAN_VERSION_CODE` to .env

### Checklist

- [x] All tests passing

### Breaking change
Remove `PostmanAuthClient`

### DIFF
--- a/app/app.py
+++ b/app/app.py
@@ -222,8 +222,7 @@ def use_route_path_as_operation_ids(app: FastAPI) -> None:
     for route in app.routes:
         if isinstance(route, APIRoute):
             # The operation_id should be unique.
-
-            # We currently never have more than one method per route
+            # It is possible to set multiple methods for the same endpoint method but it's not considered a good practice.
             method = "_".join(route.methods)
             route.operation_id = method.lower() + route.path.replace("/", "_")
 
--- a/app/app.py
+++ b/app/app.py
@@ -284,10 +284,6 @@ def get_application(settings: Settings, drop_db: bool = False) -> FastAPI:
     ):
         hyperion_error_logger.info("Redis client not configured")
 
-    if len(settings.DEPRECATED_SETTINGS) > 0:
-        for depreciation in settings.DEPRECATED_SETTINGS:
-            hyperion_error_logger.warning("DEPRECIATION NOTICE : " + depreciation)
-
     @app.middleware("http")
     async def logging_middleware(
         request: Request,
--- a/app/app.py
+++ b/app/app.py
@@ -14,6 +14,7 @@ from fastapi.encoders import jsonable_encoder
 from fastapi.exceptions import RequestValidationError
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse
+from fastapi.routing import APIRoute
 from sqlalchemy.engine import Connection, Engine
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import Session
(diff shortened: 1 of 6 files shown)


app/core/security.py
```diff
@@ -20,9 +20,9 @@
 """
 
 oauth2_scheme = OAuth2AuthorizationCodeBearer(
-    authorizationUrl="auth/authorize",
-    tokenUrl="auth/token",
-    scheme_name="Authorization Code authentication",
+    authorizationUrl="/auth/authorize",
+    tokenUrl="/auth/token",
+    scheme_name="AuthorizationCodeAuthentication",
```
armanddidierjean, line 25:
```suggestion
    authorizationUrl="/auth/authorize",
    tokenUrl="/auth/token",
    scheme_name="AuthorizationCodeAuthentication",
    scopes={"API": "Access Hyperion endpoints"},
```

app/core/config.py
```diff
@@ -101,10 +101,9 @@ class Settings(BaseSettings):
     LOG_DEBUG_MESSAGES: bool | None
     # Hyperion follows Semantic Versioning
     # https://semver.org/
-    HYPERION_VERSION: str = "1.0.0"  # This value should never be modified by hand. See [Hyperion release] documentation
-    MINIMAL_TITAN_VERSION_CODE: int = 100
-    # Depreciated, minimal_titan_version_code should be used
-    MINIMAL_TITAN_VERSION: str = "0.0.1"
+    HYPERION_VERSION: str = "0.0.1"
+    MINIMAL_TITAN_VERSION_CODE: int = 1
+    MINIMAL_TITAN_VERSION: str = "0.0.1"  # deprecated, use MINIMAL_TITAN_VERSION_CODE
```
julien4215, line 106: Why should we give a default value to these three version numbers ?

I think we want to throw an error if they are not defined in the .env

julien4215 (changes requested):

app/core/config.py
armanddidierjean: In my opinion these fields should be hardcoded and not based on the dotenv. The `MINIMAL_TITAN_VERSION_CODE` is known during development and not deployement.
Allowing to override the values in the dotenv may sometimes be usefull.

julien4215: I am not sure to understand what you mean by
known during development and not deployement

The issue I see with leaving a default value is that we might release a Hyperion without adding the versions and noticing it since it will not produce any errors

julien4215 (commented):
Why are we moving versions numbers and code to .env ? (see Armand's comment below)

> In my opinion these fields should be hardcoded and not based on the dotenv. The MINIMAL_TITAN_VERSION_CODE is known during development and not deployement.

app/core/config.py
julien4215: I am resolving this comment as the real issue seems to be that we are moving theses variable to .env

app/app.py
```diff
@@ -228,8 +236,13 @@ async def lifespan(app: FastAPI):
...
 
     # Initialize app
-    app = FastAPI(lifespan=lifespan)
+    app = FastAPI(
+        title="Hyperion",
+        version=settings.HYPERION_VERSION,
+        lifespan=lifespan,
+    )
     app.include_router(api.api_router)
+    use_route_path_as_operation_ids(app)
```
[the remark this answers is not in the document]
armanddidierjean: The old name was long and complex, as it contained the function name, the method, and the path: `create_batch_membership_groups_batch_membership_post`.

We now use `method_path`, ex: `post_groups_batch_membership`

armanddidierjean (approved):
Brilliant, thanks @Petitoto!

PR #396, opened 2024-04-04
Format and imporve Alembic migration generation

armanddidierjean:
### Description

 - import `app` in migration files for custom DATETIME objects
 - lint and format migration files
 - don't put the migration revision and down_revision in the file docstring as we usually forget to update them

### Checklist

- [x] All tests passing


Rotheem (approved):

julien4215 (commented):
Maybe we could add a comment in the datetime.py file to explain that if we change the custom type then it might break the existing migrations.

PR #400, opened 2024-04-10
Fix phonebook and CoreGroup associated with Association

Rotheem:
### Description

### Checklist

- [ ] All tests passing


armanddidierjean (changes requested):

app/modules/phonebook/cruds_phonebook.py
```diff
@@ -172,6 +172,43 @@ async def update_membership(
...
 
 
+async def update_order_of_memberships(
+    association_id: str,
+    mandate_year: int,
+    old_order: int,
+    new_order: int,
+    db: AsyncSession,
+):
+    """Update the order of Memberships in database if a Membership is moved up or down"""
```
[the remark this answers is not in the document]
Rotheem: There will never be a case where new_order is None while old_order is not as new membership are forced to be last

app/modules/phonebook/endpoints_phonebook.py
```diff
@@ -180,6 +252,14 @@ async def delete_association(
             status_code=403,
             detail=f"You are not allowed to delete association {association_id}",
         )
+    association = await cruds_phonebook.get_association_by_id(association_id, db)
+    if association is None:
+        raise HTTPException(404, "Association does not exist.")
+    if not association.deactivated:
+        raise HTTPException(
+            400,
+            "Association was not deactivated before, this is a double check system.",
```
armanddidierjean, line 261:
```suggestion
            "Only deactivated associations can be deleted.",
```

app/modules/phonebook/endpoints_phonebook.py
```diff
@@ -391,18 +494,39 @@ async def update_membership(
...
+    if updated_membership.role_tags is not None:
+        if RoleTags.president.value in updated_membership.role_tags.split(
+            ";",
+        ) and not is_user_member_of_an_allowed_group(
+            user=user,
+            allowed_groups=[GroupType.CAA, GroupType.BDE],
+        ):
+            raise HTTPException(
+                status_code=403,
+                detail="You are not allowed to update a membership with the role of president",
```
armanddidierjean, line 516:
```suggestion
                detail="Only CAA and BDE can update a membership with the role of president",
```

app/modules/phonebook/endpoints_phonebook.py
```diff
@@ -391,18 +494,39 @@ async def update_membership(
...
+            detail=f"You are not allowed to update membership for association {old_membership.association_id}",
+        )
+
+    if updated_membership.role_tags is not None:
+        if RoleTags.president.value in updated_membership.role_tags.split(
+            ";",
+        ) and not is_user_member_of_an_allowed_group(
+            user=user,
+            allowed_groups=[GroupType.CAA, GroupType.BDE],
+        ):
```
armanddidierjean, line 513: Could we create a helper function to reduce code duplication?

app/modules/phonebook/models_phonebook.py
```diff
@@ -20,20 +29,37 @@ class Membership(Base):
...
+    member_order: Mapped[int]
 
 
 class Association(Base):
     __tablename__ = "phonebook_association"
 
-    id: Mapped[str] = mapped_column(
+    id: Mapped[str] = mapped_column(String, primary_key=True, index=True)
+    name: Mapped[str] = mapped_column(String, index=True)
+    description: Mapped[str] = mapped_column(String, nullable=True)
```
armanddidierjean, line 42:
```suggestion
    description: Mapped[str | None]
```

app/modules/phonebook/models_phonebook.py
```diff
@@ -20,20 +29,37 @@ class Membership(Base):
...
+    role_tags: Mapped[str]
+    member_order: Mapped[int]
 
 
 class Association(Base):
     __tablename__ = "phonebook_association"
 
-    id: Mapped[str] = mapped_column(
+    id: Mapped[str] = mapped_column(String, primary_key=True, index=True)
+    name: Mapped[str] = mapped_column(String, index=True)
```
armanddidierjean, line 41: Do we gain from indexing `name`?

app/modules/phonebook/models_phonebook.py
```diff
@@ -20,20 +29,37 @@ class Membership(Base):
...
+    description: Mapped[str] = mapped_column(String, nullable=True)
+    kind: Mapped[Kinds]
+    mandate_year: Mapped[int]
+    deactivated: Mapped[bool]
+    associated_groups: Mapped[list["CoreGroup"]] = relationship(
+        "CoreGroup",
+        secondary="phonebook_association_associated_groups",
+        lazy="selectin",
```
armanddidierjean, line 49: Is there a benefit to the use of "selectin" instead of "joined"?

migrations/versions/16-fix_phonebook.py
```diff
@@ -0,0 +1,334 @@
...
+    #     ("3", 1),
+    #     ("4", 2),
+    #     ("5", 4),
+    #     ("6", 0),
+    #     ("7", 1),
+    #     ("8", 2),
+    # ]
+
+    # for solution in solutions:
+    #     assert solution in rows
```
armanddidierjean, line 334: We need to fix this

tests/test_phonebook.py
```diff
@@ -1,165 +1,272 @@
...
+
+association2_group2: models_phonebook.AssociationAssociatedGroups
 
-token_BDE: str
-token_president: str
-token_simple: str
+token_admin: str = ""
+token_BDE: str = ""
+token_president: str = ""
+token_simple: str = ""
```
armanddidierjean, line 40:
```suggestion
token_admin: str
token_BDE: str
token_president: str
token_simple: str
```

tests/test_phonebook.py
```diff
@@ -1,165 +1,272 @@
...
 
 
 @pytest_asyncio.fixture(scope="module", autouse=True)
-async def init_objects() -> None:
+async def init_objects(client: TestClient):
```
armanddidierjean, line 44: I think you don't need the client for this fixture

tests/test_phonebook.py
```diff
@@ -1,165 +1,272 @@
...
-    global association
-    association = models_phonebook.Association(
+    phonebook_user_simple2 = await create_user_with_groups([GroupType.student])
+    phonebook_user_simple3 = await create_user_with_groups([GroupType.student])
+
+    phonebook_user_admin = await create_user_with_groups([GroupType.admin])
+    token_admin = create_api_access_token(phonebook_user_admin)
+
+    association1_group = models_core.CoreGroup(
+        id="1",
```
armanddidierjean, line 89: Using str(UUID) would make a future migration from string to uuid type easier

armanddidierjean (changes requested):

tests/test_phonebook.py
Rotheem: UUID mess up everything, good luck

app/modules/phonebook/endpoints_phonebook.py
```diff
@@ -155,6 +177,56 @@ async def update_association(
...
+):
+    """
+    Update the groups associated with an Association
+
+    **This endpoint is only usable by CAA and BDE**
+    """
+    if not is_user_member_of_an_allowed_group(
+        user=user,
+        allowed_groups=[GroupType.CAA, GroupType.BDE],
+    ):
```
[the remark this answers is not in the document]
Rotheem: `is_user_a_member_of` allow only one group to be given, moreover we can't use it as `Depends(is_user_member_of_an_allowed_group)` as this is not the only condition that grant you access to the endpoint, presidents can also access such endpoints and it needs the `association_id` of the endpoint to be defined. This is the case only when we enter the endpoint and not during pre-check, so it has to be done this way

app/modules/phonebook/models_phonebook.py
Rotheem: I just copied the way it was done in reception's PR

app/modules/phonebook/models_phonebook.py
```diff
@@ -20,20 +29,37 @@ class Membership(Base):
...
 class Association(Base):
     __tablename__ = "phonebook_association"
 
-    id: Mapped[str] = mapped_column(
+    id: Mapped[str] = mapped_column(String, primary_key=True, index=True)
+    name: Mapped[str] = mapped_column(String, index=True)
+    description: Mapped[str] = mapped_column(String, nullable=True)
+    kind: Mapped[Kinds]
+    mandate_year: Mapped[int]
+    deactivated: Mapped[bool]
```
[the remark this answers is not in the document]
Rotheem: Done

armanddidierjean (approved):

PR #406, opened 2024-04-19
Separate production and developement requirements

julien4215:
- Production and development requirements are now separated
- Switch from psycopg2 to psycopg3
- Use pscyopg3 binary installation for dev and local installation for prod


armanddidierjean (approved):
Works great!

Maybe just remove the `.dockerignore`

PR #409, opened 2024-04-21
Database and models overall improvements

julien4215:
- remove version in docker compose
- keep unused import during alembic hook
- configuration to use annotated declarative table


guillaumetavernier (approved):

app/types/sqlalchemy.py
```diff
@@ -27,3 +30,21 @@ def process_result_value(self, value, dialect):
         if value is not None:
             value = value.replace(tzinfo=datetime.UTC)
         return value
+
+
+# Pre-configured field type for UUID primary key (see https://docs.sqlalchemy.org/en/20/orm/declarative_tables.html#mapping-whole-column-declarations-to-python-types)
+primary_key = Annotated[uuid.UUID, mapped_column(primary_key=True)]
```
armanddidierjean, line 36: I think it may be a better practice to use CamelCase for this type.
See https://peps.python.org/pep-0008/#type-variable-names
```suggestion
PrimaryKey = Annotated[uuid.UUID, mapped_column(primary_key=True)]
```

pyproject.toml
```diff
@@ -50,6 +50,7 @@ ignore = [
     "SIM102",
     "SIM105",
     "RUF012", # We may want to enable "Mutable class attributes should be annotated with `typing.ClassVar`"
+    "COM812", # Not recommended because we also use ruff as a formatter (see https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules)
```
armanddidierjean, line 53: Could we move this change to an other PR, as it is unrelated with the other changes about SQLAlchemy and database.

To be honest, I would be in favor of keeping this rule, which requires trailing comma. In practice it seems to work well with the formater.

armanddidierjean (changes requested):
Is there a reason to change the type the recommandation module only?

julien4215:
> Is there a reason to change the type the recommandation module only?

No particular reason, I just chose this module because it is a simple one and I wanted to give an example on how to use annotated declarative table and migrate primary key type from string to uuid. We should try to move all code to this new syntax but it will require a lot of work.

pyproject.toml
julien4215: > Could we move this change to an other PR, as it is unrelated with the other changes about SQLAlchemy and database.
Sure, I added this change here because I consider this pull request as several individual improvement of the code but I will move it to a different PR as it is more controversial than I expected.

migrations/versions/8-recommendation_primary_key.py
```diff
@@ -0,0 +1,32 @@
...
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.alter_column(
+        "recommendation",
+        "id",
+        type_=sa.types.Uuid(),
+        postgresql_using="id::uuid",
+    )
```
armanddidierjean, line 24: To be able to use it with SQLite, could you wrap alters in a [batch migration bloc](https://alembic.sqlalchemy.org/en/latest/batch.html) ?

```suggestion
    with op.batch_alter_table("recommendation") as batch_op:
        batch_op.alter_column(
            "id",
            type_=sa.types.Uuid(),
            postgresql_using="id::uuid",
        )
```

armanddidierjean (changes requested):

alembic.ini
```diff
@@ -71,7 +71,7 @@ hooks=ruff_check, ruff_format
 
 ruff_check.type = exec
 ruff_check.executable = ruff
-ruff_check.options = check --fix REVISION_SCRIPT_FILENAME
+ruff_check.options = check --ignore F401 --fix REVISION_SCRIPT_FILENAME
```
armanddidierjean, line 74: This addition is really useful!
Could you maybe add a comment explaining why we don't want missing imports to be removed during the migration generation?

armanddidierjean (approved):
Works great with SQLite

PR #410, opened 2024-04-22
Ignore COM812 rule

julien4215:
Remove a rule that is conflicting with the formatter


Rotheem (approved):

armanddidierjean:
Related issue: https://github.com/astral-sh/ruff/issues/9216

Personally, I would be in favor of keeping COM812, even if it's tagged as conflicting with the formatter

PR #412, opened 2024-04-26
Topic notification transition

### LINKED ISSUE #291: Planed notifications bug
### Subject of the issue

Planed notifications (eg: cinema recap) don't always work correctly:
- they are not triggered at the end of the month (see below)
- sometimes, more than one notification are triggered (no explanation for this bug yet)

In cinema endpoints, this part of the code does not work if the next sunday is in the next month:
```
today = datetime.now(timezone.utc)
sunday = today.replace(
    day=today.day - today.weekday() + 6, hour=11, minute=0, second=0
)
next_week_sessions = await cruds_cinema.get_sessions_in_time_frame(
    start_after=sunday, start_before=sunday.replace(day=sunday.day + 7), db=db
)
```

### Your environment

- Version of Hyperion: latest release
- Operating system: Android

### Steps to reproduce

Register to cinema notifications and wait for next sunday.

### Expected behaviour

Get exactly one notification at the right time.

### Actual behaviour

Sometimes notifications aren't triggered, sometimes multiple notifications are triggered.

Foukki:
### Description

Sending messages to a topic doesn't work on iOS when the app is in background. This branch will fix this by sending the messages to the tokens of users subscribed to the topic, instead of sending them to the topic directly. We can send a message to 500 devices per request, so in most cases we won't need to make more than one request to firebase.

The PR add tests for notifications. Due to the usage of Firebase messaging to send notifications, tests can not check notifications can be sent. Instead, they make sure our endpoints to register/unregister a device or subscribe to topics work.

To prevent anyone from subscribing to `bookingadmin` topic and then be notified of all booking requests, the notification is now sent to all members of the room manager group. We way want to send the notification to just a few persons instead of the whole group.

Fix #291


app/core/notification/endpoints_notification.py
```diff
@@ -298,6 +298,15 @@ async def send_notification(
 
     **Only admins can use this endpoint**
     """
+    now = datetime.now(UTC)
+    message = Message(
```
armanddidierjean, line 302:
```suggestion
    message = schemas_notification.Message(
```

armanddidierjean (changes requested):

app/modules/loan/endpoints_loan.py
```diff
@@ -599,6 +614,43 @@ async def create_loan(
...
+        hyperion_error_logger.error(
+            f"Error while sending notification to borrower of a new loan, {error}",
+        )
+
+    try:
+        now = datetime.now(UTC)
+        message = Message(
+            context=f"loan-new-{loan.id}-end-notif",
+            is_visible=True,
+            title="📦 Prêt arrivé à échéance",
```
armanddidierjean, line 641: What happen if the loan is returned before the due date, or if the date is modified?

app/core/notification/endpoints_notification.py
```diff
@@ -298,6 +297,15 @@ async def send_notification(
...
     **Only admins can use this endpoint**
     """
+    now = datetime.now(UTC)
+    message = schemas_notification.Message(
+        context="Test",
+        is_visible=True,
+        title="Test notification",
+        content="Ceci est un test de notification",
+        # The notification will expire in 3 days
+        expire_on=now.replace(day=now.day + 3),
```
armanddidierjean, line 307:
```suggestion
        expire_on=datetime.now(UTC) + timedelta(days=3),
```

app/modules/advert/endpoints_advert.py
```diff
@@ -268,19 +268,21 @@ async def create_advert(
     try:
         now = datetime.now(UTC)
         message = Message(
-            context=f"advert-{result.id}",
+            context=f"new-advert-{id}",
```
armanddidierjean, line 271: I would be in favor of always prefixing the context by the module name to avoid conflicts between modules.
```suggestion
            context=f"advert-new-{id}",
```

app/modules/booking/endpoints_booking.py
```diff
@@ -265,26 +266,30 @@ async def create_booking(
...
-                # We use sunday date as context to avoid sending the recap twice
-                context=f"booking-create-{result.id}",
+                context=f"new-booking-{id}",
                 is_visible=True,
-                title="Réservations - Nouvelle réservation 📅",
+                title="📅 Réservations - Nouvelle réservation ",
                 content=f"{result.applicant.nickname} - {result.room.name} {result.start.strftime('%m/%d/%Y, %H:%M')} - {result.reason}",
-                # The notification will expire the next sunday
+                # The notification will expire in 3 days
                 expire_on=now.replace(day=now.day + 3),
```
armanddidierjean, line 283:
```suggestion
                expire_on=datetime.now(UTC) + timedelta(days=3),
```

app/modules/cinema/endpoints_cinema.py
```diff
@@ -64,44 +65,50 @@ async def create_session(
...
+            "Avril",
+            "Mai",
+            "Juin",
+            "Juillet",
+            "Août",
+            "Septembre",
+            "Octobre",
+            "Novembre",
+            "Décembre",
+        ]
```
armanddidierjean, line 83: Could you create a util to get the current localized day and month?

app/modules/cinema/endpoints_cinema.py
```diff
@@ -64,44 +65,50 @@ async def create_session(
...
+        french_hour = db_session.start.astimezone(ZoneInfo("Europe/Paris"))
```
armanddidierjean, line 84:
```suggestion
        # As the notification content is set in the back, the front won't be able to display it in the local timezone
        # We decide to use UTC+2 for all users
        french_hour = db_session.start.astimezone(ZoneInfo("Europe/Paris"))
```

app/utils/communication/notifications.py
```diff
@@ -292,51 +242,35 @@ async def send_notification_to_user(
...
+        user_ids: list[str],
+        message: Message,
+        db: AsyncSession,
+    ) -> None:
+        for user_id in user_ids:
+            await self.send_notification_to_user(
+                user_id=user_id,
+                message=message,
+                db=db,
+            )
```
armanddidierjean, line 256: Here you will make one firebase request per user. You may prefer to get all firebase tokens for your users and then call:
```python
await self._send_firebase_trigger_notification_by_tokens(
                tokens=firebase_device_tokens,
                db=db,
            )
```
---
`send_notification_to_user(user_id: str)` could call `send_notification_to_users(user_ids=[user_id])`

armanddidierjean (changes requested):

app/utils/communication/notifications.py
Foukki: I agree with the first part but I don't really understand why send_notification_to_user(user_id: str) should call send_notification_to_users(user_ids=[user_id]) ?

armanddidierjean: I wouldn't say it should, but it may be a way to reduce code duplication between `send_notification_to_user` and `send_notification_to_users`

Foukki: I don't see any code duplication (if you meant in `notification.py`

Foukki: I don't see any code duplication (if you meant in `notification.py`)

Foukki: I don't see any code duplication (if you meant in `notification.py`)

app/modules/amap/endpoints_amap.py
```diff
@@ -718,6 +720,40 @@ async def open_ordering_of_delivery(
         )
 
     await cruds_amap.open_ordering_of_delivery(delivery_id=delivery_id, db=db)
+    try:
```
armanddidierjean, line 723: Do we keep the try/except ?

app/modules/booking/endpoints_booking.py
```diff
@@ -1,15 +1,19 @@
...
-from datetime import UTC, datetime
+from datetime import UTC, datetime, timedelta
+from zoneinfo import ZoneInfo
 
 from fastapi import Depends, HTTPException
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from app.core import models_core
+
+# from app.core.config import Settings
```
armanddidierjean, line 11:
```suggestion
```

app/modules/booking/endpoints_booking.py
```diff
@@ -265,26 +269,45 @@ async def create_booking(
...
+        manager_group_id = result.room.manager_id
+        manager_group = await cruds_groups.get_group_by_id(
+            db=db,
+            group_id=manager_group_id,
+        )
+
+        applicant_user = await cruds_users.get_user_by_id(
+            db=db,
+            user_id=result.applicant_id,
+        )
```
armanddidierjean, line 282: I think the applicant should always be the user making the request

app/modules/booking/endpoints_booking.py
```diff
@@ -265,26 +269,45 @@ async def create_booking(
...
+        if applicant_user:
+            if applicant_user.nickname:
+                applicant_nickname = applicant_user.nickname
+            else:
+                applicant_nickname = applicant_user.firstname
+            content = f"{applicant_nickname} - {result.room.name} {result.start.strftime('%m/%d/%Y, %H:%M')} - {result.reason}"
```
armanddidierjean, line 288:
```suggestion
            content = f"{applicant_user.nickname or applicant_user.firstname} - {result.room.name} {result.start.strftime('%m/%d/%Y, %H:%M')} - {result.reason}"
```

app/modules/loan/endpoints_loan.py
```diff
@@ -599,6 +615,43 @@ async def create_loan(
...
+            content=f"N'oublie pas de rendre ton prêt à l'association {loan.loaner.name} ! ",
+            delivery_datetime=delivery_datetime,
+            expire_on=loan.end + timedelta(days=30),
+        )
+
+        await notification_tool.send_notification_to_users(
+            user_ids=[loan.borrower_id],
+            message=message,
+        )
```
armanddidierjean, line 649:
```suggestion
        await notification_tool.send_notification_to_user(
            user_id=loan.borrower_id,
            message=message,
        )
```

app/modules/loan/endpoints_loan.py
```diff
@@ -599,6 +615,43 @@ async def create_loan(
...
+            f"Error while sending notification to borrower of a new loan, {error}",
+        )
+    try:
+        delivery_time = time(11, 00, 00)
+        delivery_datetime = datetime.combine(loan.end, delivery_time)
+        message = Message(
+            context=f"loan-new-{loan.id}-end-notif",
+            is_visible=True,
+            title="📦 Prêt arrivé à échéance",
+            content=f"N'oublie pas de rendre ton prêt à l'association {loan.loaner.name} ! ",
```
armanddidierjean, line 641:
```suggestion
            content=f"N'oublie pas de rendre ton prêt à l'association {loan.loaner.name} !",
```

app/modules/loan/endpoints_loan.py
```diff
@@ -828,6 +882,40 @@ async def return_loan(
...
+            user_ids=[loan.borrower_id],
+            db=db,
+        )
+        if (
+            await cruds_notification.get_messages_by_context_and_firebase_tokens(
+                context=f"loan-new-{loan.id}-end-notif",
+                firebase_tokens=device_tokens,
+                db=db,
+            )
+            != []
```
armanddidierjean, line 896: If a user has two devices, a message may still be in database for a device, but already delivered for the other. I think you always need to send the notification.

armanddidierjean (changes requested):
You can remove the topic subscription logic in `endpoints_notification`.
I'm not sure if it's still useful to send notification from try/except blocs.

app/modules/booking/endpoints_booking.py
Foukki: I don't really understant what you mean here.

armanddidierjean (approved):

maximeroucher (approved):
Looks good !

PR #413, opened 2024-04-27
Annotations for dependecies

julien4215:
https://fastapi.tiangolo.com/release-notes/#0950


app/dependencies.py
```diff
@@ -274,9 +269,12 @@ def is_user_a_member(
     return user
 
 
+MemberUser = Annotated[models_core.CoreUser, Depends(is_user_a_member)]
```
guillaumetavernier, line 272: Shouldn't it be `UserMember` to stay consistant with the rest of the code ?

PR #444, opened 2024-05-19
Use uuid type for the new modules

julien4215:


app/modules/ph/models_ph.py
```diff
@@ -1,14 +1,13 @@
...
 
 
 class Paper(Base):
     __tablename__ = "ph_papers"
 
-    id: Mapped[str] = mapped_column(String, primary_key=True, index=True)
-    name: Mapped[str] = mapped_column(String, nullable=False)
-    release_date: Mapped[date] = mapped_column(Date, nullable=False)
+    id: Mapped[PrimaryKey]
+    name: Mapped[str]
```
guillaumetavernier, line 12: Is it nullable or did the default value changed ?

julien4215: SQL column is automatically not nullable if the python type is not nullable.

guillaumetavernier (approved):

Foukki (approved):

armanddidierjean (changes requested):
I would be in favor of creating a new migration. We could maybe use this PR to add PrimaryKeys to all modules

armanddidierjean (approved):

PR #455, opened 2024-05-29
fix : leaderboard

Foukki:


app/modules/flappybird/endpoints_flappybird.py
```diff
@@ -104,12 +104,36 @@ async def create_flappybird_score(
...
+        user_id=user.id,
+        value=flappybird_score.value,
+        creation_time=creation_time,
+    )
+    personal_best = await cruds_flappybird.get_flappybird_personal_best_by_user_id(
+        user_id=user.id,
+        db=db,
+    )
+    if not personal_best:
+        try:
```
armanddidierjean, line 119: I'm not sure try/except is really useful as they don't do anything apart from raising a new exception which won't be logged. I think you could remove them

app/modules/flappybird/schemas_flappybird.py
```diff
@@ -20,6 +20,10 @@ class FlappyBirdScoreInDB(FlappyBirdScore):
     user_id: str
 
 
+class FlappyBirdBestScoreInDB(FlappyBirdScoreInDB):
+    pass
```
armanddidierjean, line 24: Is this schema really useful?

app/modules/flappybird/endpoints_flappybird.py
```diff
@@ -104,12 +104,36 @@ async def create_flappybird_score(
...
+            await cruds_flappybird.create_flappybird_best_score(
```
armanddidierjean, line 120: You may want a test to cover this line

app/modules/flappybird/cruds_flappybird.py
```diff
@@ -92,3 +67,32 @@ async def create_flappybird_score(
...
+        await db.commit()
+        return flappybird_best_score
+    except IntegrityError as error:
+        await db.rollback()
+        raise ValueError(error)
+
+
+async def update_flappybird_best_score(
+    db: AsyncSession,
+    flappybird_best_score: models_flappybird.FlappyBirdBestScore,
```
armanddidierjean, line 88: I think you could accept `user_id` and `best_score_value` directly

armanddidierjean (changes requested):

armanddidierjean (approved):

BLEMENT33 (approved):

PR #471, opened 2024-06-11
Cinema notif on current week film

Foukki:
### Description

If we add a film this week the release date of the notification is last sunday so it triggers notification now. That is not what we want.

### Checklist

- [ x ] All tests passing


armanddidierjean (changes requested):
Do you really need to send a notification if it won't be displayed nor used by Titan?

armanddidierjean (approved):

PR #477, opened 2024-06-15
Fix loggers and db initialization when using multiple Gunicorn workers

armanddidierjean:
### Description

- disable default Gunicorn logger handlers to use our custom loggers
- remove `--preload` flag which break our async Queue logger handlers
- allow to prevent db init on Hyperion startup using an env variable
- add a Gunicorn on_starting hook to init the db

### Checklist

- [ ] All tests passing

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,7 +205,7 @@ To enable the service:
 
 ## Running Hyperion with Gunicorn
 
-For production we encourage to run Hyperion with multiple Gunicorn workers. You can use our [docker image](./Dockerfile) and [docker-compose file](./docker-compose.yaml) files to run Hyperion with Gunicorn.
+For production we encourage to use Gunicorn to run and manage multiple Uvicorn workers. You can use our [docker image](./Dockerfile) and [docker-compose file](./docker-compose.yaml) files to run Hyperion with Gunicorn. See [Gunicorn with Uvicorn](https://fastapi.tiangolo.com/deployment/server-workers/#gunicorn-with-uvicorn-workers) FastAPI documentation.
 
 Do not use gunicorn `--preload` flag. It initialise a first Hyperion instance then fork it to create workers. This is not compatible with the way we handle loggers in their own thread.
 
--- a/README.md
+++ b/README.md
@@ -200,3 +200,13 @@ To enable the service:
    1. Go to [Google cloud, IAM and administration, Service account](https://console.cloud.google.com/iam-admin/serviceaccounts) and add a new Service Account with Messaging API capabilities.
    2. Choose _Manage keys_ and create a new JSON key.
    3. Rename the file `firebase.json` and add it at Hyperion root
+
+---
+
+## Running Hyperion with Gunicorn
+
+For production we encourage to run Hyperion with multiple Gunicorn workers. You can use our [docker image](./Dockerfile) and [docker-compose file](./docker-compose.yaml) files to run Hyperion with Gunicorn.
+
+Do not use gunicorn `--preload` flag. It initialise a first Hyperion instance then fork it to create workers. This is not compatible with the way we handle loggers in their own thread.
+
+You should use our [Gunicorn configuration file](./gunicorn_conf.py) to ensure that database initialization and migrations are only run once.
--- a/app/core/config.py
+++ b/app/core/config.py
@@ -176,6 +176,11 @@ class Settings(BaseSettings):
     # Automatically generated parameters #
     ######################################
 
+    # If Hyperion should initialize the database on startup
+    # This environment variable is set by the Gunicorn on_starting hook, to tell the workers to avoid initializing the database
+    # You don't want to set this variable manually
+    HYPERION_INIT_DB: bool = True
+
     # The following properties can not be instantiated as class variables as them need to be computed using another property from the class,
     # which won't be available before the .env file parsing.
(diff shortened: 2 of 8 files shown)


app/core/config.py
```diff
@@ -157,10 +150,29 @@ class Settings(BaseSettings):
...
 
+    # If Hyperion should initialize the database on startup
+    # This environment variable is set by the Gunicorn on_starting hook, to tell the workers to avoid initializing the database
+    # You don't want to set this variable manually
+    HYPERION_INIT_DB: bool = True
+
+    # An old variable used by clients to determine if they are compatible with Hyperion
+    # Using 0.0.1 as the minimal version, force old and broken clients to show a message asking to update
+    # TODO: remove this property when old clients are updated
+    MINIMAL_TITAN_VERSION: str = "0.0.1"  # deprecated, use MINIMAL_TITAN_VERSION_CODE
```
julien4215, line 174: Maybe it is time to remove this property ?

If we keep it, why did you put it in automatically generated parameters ?

app/core/log.py
```diff
@@ -220,6 +220,18 @@ def get_config_dict(self, settings: Settings):
...
+                "gunicorn.access": {"handlers": []},
+                "gunicorn.error": {
+                    "handlers": [
+                        "file_errors",
+                        "matrix_errors",
+                        "console",
+                    ],
+                    "level": MINIMUM_LOG_LEVEL,
+                    "propagate": False,
+                },
```
julien4215, line 234: I think this should be moved in a different PR. Maybe in this one #452 ?

julien4215 (commented):
Amazing PR ! I did not have time to test it yet but I'll do it later.

We should add gunicorn as a dev dependency.

README.md
```diff
@@ -200,3 +200,13 @@ To enable the service:
    1. Go to [Google cloud, IAM and administration, Service account](https://console.cloud.google.com/iam-admin/serviceaccounts) and add a new Service Account with Messaging API capabilities.
    2. Choose _Manage keys_ and create a new JSON key.
    3. Rename the file `firebase.json` and add it at Hyperion root
+
+---
+
+## Running Hyperion with Gunicorn
+
+For production we encourage to run Hyperion with multiple Gunicorn workers. You can use our [docker image](./Dockerfile) and [docker-compose file](./docker-compose.yaml) files to run Hyperion with Gunicorn.
```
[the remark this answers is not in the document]
armanddidierjean: I think we can use Gunicorn with Uvicorn workers, but I'm not sure we use that in production.

What would you think of:
```suggestion
For production we encourage to run Hyperion with Gunicorn and multiple workers. You can use our [docker image](./Dockerfile) and [docker-compose file](./docker-compose.yaml) files to run Hyperion with Gunicorn.
```

app/core/config.py
armanddidierjean: Good idea, I will make a second PR

README.md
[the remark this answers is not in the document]
julien4215: I am pretty sure that the image we use run multiple Uvicorn workers with Gunicorn so I would suggest
```suggestion
For production we encourage to use Gunicorn to run and manage multiple Uvicorn workers. You can use our [docker image](./Dockerfile) and [docker-compose file](./docker-compose.yaml) files to run Hyperion with Gunicorn.
```

Maybe we could simply add a link to the FastAPI documentation https://fastapi.tiangolo.com/deployment/server-workers/#gunicorn-with-uvicorn-workers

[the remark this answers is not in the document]
armanddidierjean: You're completely right, mb.
```suggestion
For production we encourage to use Gunicorn to run and manage multiple Uvicorn workers. You can use our [docker image](./Dockerfile) and [docker-compose file](./docker-compose.yaml) files to run Hyperion with Gunicorn. See [Gunicorn with Uvicorn](https://fastapi.tiangolo.com/deployment/server-workers/#gunicorn-with-uvicorn-workers) FastAPI documentation.
```

armanddidierjean:
> Don't we need to add gunicorn as a dev dependency ?
If we continue to use uvicorn as development server, I would say we don't need to add gunicorn as a dependency.

gunicorn.conf.py
```diff
@@ -0,0 +1,34 @@
+import logging
+import os
+
+from gunicorn.arbiter import Arbiter
```
julien4215, line 4: We use gunicorn here and it will not be recognized if we don't add it to dev requirements

armanddidierjean: Is it really an issue? The CI doesn't complain, if we don't install and use Gunicorn this file won't be used.

julien4215: The CI doesn't complain because ruff still didn't implement E0401 rule.

I think we should give to developers every modules we use in our code.

julien4215: Actually do we need to import the Arbiter class ? I believe we do not need to type `on_starting` `server` argument

armanddidierjean: We don't really, the objective was to type the method, but we don't use the `server` argument. I will just add a comment

pyproject.toml
```diff
@@ -108,7 +108,9 @@ disallow_any_generics = false
 disallow_untyped_calls = false
 
 [[tool.mypy.overrides]]
-module = ["firebase_admin", "icalendar", "fitz"]
+# Tracking issues
+# Gunicorn : https://github.com/benoitc/gunicorn/issues/2833
+module = ["firebase_admin", "icalendar", "fitz", "gunicorn.arbiter"]
```
julien4215, line 113: I think we can remove this too

Rotheem (approved):

Malem38 (approved):

PR #491, opened 2024-07-07
Use jaro_winkler similarity instead of rapidfuzz

Rotheem:
### Description

Use jaro_winkler similarity from jellyfish module instead of rapidfuzz. It has the same speed and better results

### DIFF
--- a/app/utils/tools.py
+++ b/app/utils/tools.py
@@ -82,9 +82,17 @@ def sort_user(
         for user, choice in zip(users, names, strict=True)
     ]
 
-    scored.sort(key=lambda x: x[1], reverse=True)
+    results = []
+    for _ in range(10):
+        results.append(
+            next(
+                result
+                for result in scored
+                if result[1] == max(scored, key=lambda r: r[1])[1]
+            ),
+        )
 
-    return [user for user, _ in scored[:limit]]
+    return [result[0] for result in results]
 
 
 async def is_group_id_valid(group_id: str, db: AsyncSession) -> bool:
--- a/app/utils/tools.py
+++ b/app/utils/tools.py
@@ -84,13 +84,9 @@ def sort_user(
 
     results = []
     for _ in range(10):
-        results.append(
-            next(
-                result
-                for result in scored
-                if result[1] == max(scored, key=lambda r: r[1])[1]
-            ),
-        )
+        maximum = max(scored, key=lambda r: r[1])
+        results.append(maximum)
+        scored.remove(maximum)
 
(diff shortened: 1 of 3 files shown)


app/utils/tools.py
```diff
@@ -71,26 +71,20 @@ def fuzzy_search_user(
...
     The size of the answer can be limited using `limit` parameter.
 
-    Use RapidFuzz library
+    Use Jellyfish library
     """
 
-    # We can give a dictionary of {object: string used for the comparison} to the extract function
-    # https://maxbachmann.github.io/RapidFuzz/Usage/process.html#extract
-
     # TODO: we may want to cache this object. Its generation may take some time if there is a big user base
```
maximeroucher, line 77: Isn't it the time to do it ?

app/utils/tools.py
```diff
@@ -71,26 +71,20 @@ def fuzzy_search_user(
...
-    results: list[tuple[str, int | float, int]] = process.extract(
-        query,
-        choices,
-        limit=limit,
-    )
+    scored.sort(key=lambda x: x[1], reverse=True)
 
-    # results has the format : (string used for the comparison, similarity score, index of the object in the choices collection)
-    return [users[res[2]] for res in results]
+    return [user for user, _ in scored[:limit]]
```
armanddidierjean, line 87: I'm not sure what could be a good way to sort only the first `limit` elements. Currently, we sort 1000 users to return only the first 10

app/utils/tools.py
armanddidierjean: Instead of caching we may explore the list generation in the database
https://docs.sqlalchemy.org/en/20/core/functions.html#sqlalchemy.sql.functions.concat

app/utils/tools.py
Rotheem: The new system should be better

app/utils/tools.py
```diff
@@ -71,26 +71,28 @@ def fuzzy_search_user(
...
+        (user, jaro_winkler_similarity(query, choice))
+        for user, choice in zip(users, names, strict=True)
+    ]
+
+    results = []
+    for _ in range(10):
+        results.append(
+            next(
+                result
+                for result in scored
```
armanddidierjean, line 90:
```suggestion
                for user, score in scored
```

armanddidierjean (changes requested):

armanddidierjean (approved):
Nice!

Malem38 (approved):

PR #494, opened 2024-07-18
Add colors to different levels of log in console

Rotheem:
Add a custom formatter used by console's logs to addapt the color depending on log's level


app/core/log.py
```diff
@@ -1,13 +1,45 @@
...
 from logging.handlers import QueueHandler, QueueListener
 from pathlib import Path
 from typing import Any
 
+import uvicorn
+
 from app.core.config import Settings
 
 
+class ModifiedFormatter(uvicorn.logging.DefaultFormatter):
```
armanddidierjean, line 14: Maybe `ColoredConsoleFormatter` ?
```suggestion
class ColoredConsoleFormatter(uvicorn.logging.DefaultFormatter):
```

app/core/log.py
```diff
@@ -1,13 +1,45 @@
...
+    class console_color(Enum):
```
armanddidierjean, line 15: I *think* if `console_color` inherit from str, you won't need to call `.value` in strings.

armanddidierjean, line 15:
```suggestion
    class ConsoleColor(Enum):
```

armanddidierjean (changes requested):

app/core/log.py
```diff
@@ -1,13 +1,45 @@
...
+            "%(asctime)s - %(name)s - "
+            + self.console_color.BOLD.value
+            + "%(levelname)s"
+            + self.console_color.END.value
+            + " - "
+            + self.console_color[record.levelname].value
+            + "%(message)s"
+            + self.console_color.END.value
+        )
+        formatter = logging.Formatter(fmt, self.datefmt)
```
armanddidierjean, line 39: I'm not sure if this is really pertinent, but could we instantiate `logging.Formatter` instead of creating a new one for each `format()` call?

Rotheem: The original solution was calling the super().format() method but even when setting self._fmt it wasn't able to use the modified format. That's the solution I found on [Stack Overflow](https://stackoverflow.com/questions/384076/how-can-i-color-python-logging-output) that allow to color only a certain part of the text

requirements-dev.txt
```diff
@@ -11,4 +11,5 @@ pytest==8.2.2
 ruff==0.4.8
 types-aiofiles==24.1.0.20240626
 types-redis==4.6.0.20240425
-types-requests==2.32.0.20240622
\ No newline at end of file
+types-requests==2.32.0.20240622
+types-zxcvbn==4.4.1.20240106
```
armanddidierjean, line 15: This change should be in #377

armanddidierjean (changes requested):

app/core/log.py
```diff
@@ -1,13 +1,46 @@
...
+            + self.ConsoleColors.BOLD
+            + "%(levelname)s"
+            + self.ConsoleColors.END
+            + " - "
+            + self.ConsoleColors[record.levelname].value
+            + "%(message)s"
+            + self.ConsoleColors.END
+        )
+        formatter = logging.Formatter(fmt, self.datefmt)
+        return formatter.format(record)
```
armanddidierjean, line 41: I'm not sure about this idea, but could this be better to prevent a new formatter instantiation for each `format` call?
```suggestion
    class ConsoleColors(str, Enum):
        """Colors can be found here: https://talyian.github.io/ansicolors/"""

        DEBUG = "\033[38;5;12m"
        INFO = "\033[38;5;10m"
        WARNING = "\033[38;5;11m"
        ERROR = "\033[38;5;9m"
        CRITICAL = "\033[38;5;1m"
        BOLD = "\033[1m"
        END = "\033[0m"

    def __init__(self, *args, **kwargs):
        super().__init__(datefmt="%d-%b-%y %H:%M:%S")
        
        self.formatters = {}
        
        for level in [logging.DEBUG, logging.INFO, logging.WARNING, logging.ERROR, logging.CRITICAL]:
            fmt = (
                "%(asctime)s - %(name)s - "
                + self.ConsoleColors.BOLD
                + "%(levelname)s"
                + self.ConsoleColors.END
                + " - "
                + self.ConsoleColors[level]
                + "%(message)s"
                + self.ConsoleColors.END
            )
            level_formatter = logging.Formatter(fmt, self.datefmt)
            self.formatters[level] = level_formatter

    def format(self, record: logging.LogRecord) -> str:
        formatter = self.formatters.get(record.levelno, self.formatters[logging.ERROR])
        return formatter.format(record)
```

requirements-dev.txt
Rotheem: Hum my bad

armanddidierjean (approved):
Lovely

PR #498, opened 2024-07-30
Unit of work

Rotheem:
### Description

### Checklist

- [ ] All tests passing

### DIFF
--- a/app/dependencies.py
+++ b/app/dependencies.py
@@ -130,7 +130,6 @@ async def get_transactional_db() -> AsyncGenerator[TransactionalAsyncSession, No
 
     See TransactionalAsyncSession definition for more informations
     """
-    global SessionLocal
     if SessionLocal is None:
         hyperion_error_logger.error("Database engine is not initialized")
         raise HTTPException(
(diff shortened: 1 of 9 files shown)


app/dependencies.py
```diff
@@ -86,9 +88,12 @@ def init_and_get_db_engine(settings: Settings) -> AsyncEngine:
         SQLALCHEMY_DATABASE_URL = f"postgresql+asyncpg://{settings.POSTGRES_USER}:{settings.POSTGRES_PASSWORD}@{settings.POSTGRES_HOST}/{settings.POSTGRES_DB}"
 
     if engine is None:
+        # The use of transactional database requires the use of NullPool as it implies the use of a db in multiple run loops
+        # https://docs.sqlalchemy.org/en/14/orm/extensions/asyncio.html#using-multiple-asyncio-event-loops
```
armanddidierjean, line 92: We need to clear the usage of NullPool up. This should not break tests as this dependency should not be called during tests

armanddidierjean (changes requested):

PR #516, opened 2024-08-14
Add Documenso provider

armanddidierjean:


Rotheem (approved):

maximeroucher (approved):

migrations/versions/25-hardcode-eclair-group.py
```diff
@@ -0,0 +1,151 @@
...
+
+    # We update relationships to use the new group id
+    conn.execute(
+        t_membership.update()
+        .where(t_membership.c.group_id == old_eclair_id)
+        .values(group_id=new_eclair_id),
+    )
+
+    # We delete the old group
+    conn.execute(t_group.delete().where(t_group.c.id == old_eclair_id))
```
Rotheem, line 77: This will fail as eclair's group is used in seller, loan...

Rotheem, line 77: This will fail as eclair's group is used in seller, loan,...

Rotheem (changes requested):

Rotheem (approved):

PR #544, opened 2024-08-26
CDR : add multiple tickets generation

Malem38:
### Description

### Checklist

- [ ] All tests passing


maximeroucher (approved):

Rotheem (approved):

app/modules/cdr/endpoints_cdr.py
```diff
@@ -351,6 +355,197 @@ async def get_online_sellers(
...
+        file_path,
+        index=False,
+        freeze_panes=(2, 3),
+        engine="xlsxwriter",
+    )
+    send_email(
+        recipient=email.email,
+        subject=f"Résultats de ventes pour {seller.name}",
+        content=f"Bonjour,\n\nVous trouverez en pièce jointe le fichier Excel contenant les résultats de ventes pour la CdR pour l'association {seller.name}.",
+        settings=get_settings(),
```
armanddidierjean, line 541: Can not call `get_settings` directly

app/modules/cdr/endpoints_cdr.py
```diff
@@ -351,6 +355,197 @@ async def get_online_sellers(
...
+        users_answers=users_answers,
+    )
+    file_path = f"/tmp/CdR {datetime.now(tz=UTC).year} ventes {seller.name}.xlsx"  # noqa: S108
+    df.to_excel(
+        file_path,
+        index=False,
+        freeze_panes=(2, 3),
+        engine="xlsxwriter",
+    )
+    send_email(
```
armanddidierjean, line 537: You should always call `send_email` in a background task

app/modules/cdr/endpoints_cdr.py
```diff
@@ -351,6 +355,197 @@ async def get_online_sellers(
...
+        )
+    df = construct_dataframe_from_users_purchases(
+        users_purchases=purchases_by_users,
+        users=list(users),
+        products=list(products),
+        variants=variants,
+        data_fields=product_fields,
+        users_answers=users_answers,
+    )
+    file_path = f"/tmp/CdR {datetime.now(tz=UTC).year} ventes {seller.name}.xlsx"  # noqa: S108
```
armanddidierjean, line 530: I'm not completely sure if it would work, but S108 looks interesting.

armanddidierjean: If we don't use a temp file as proposed by S108, please place this file in `/data/...` which is gitignored.
Always use file manipulation utils to access the file system, to prevent injections in filepath

app/utils/mail/mailworker.py
```diff
@@ -1,13 +1,22 @@
...
     from app.core.config import Settings
 
 
-def send_email(recipient: str, subject: str, content: str, settings: "Settings"):
+def send_email(
+    recipient: str | list[str],
+    subject: str,
+    content: str,
+    settings: "Settings",
+    file_path: str | None = None,
```
armanddidierjean, line 16: Ask for folder name and file uuid, then use file manipulation utils, to prevent access to files outside `/data`

app/utils/mail/mailworker.py
```diff
@@ -1,13 +1,22 @@
...
+from pathlib import Path
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
     from app.core.config import Settings
 
 
-def send_email(recipient: str, subject: str, content: str, settings: "Settings"):
+def send_email(
+    recipient: str | list[str],
```
armanddidierjean, line 12: I think we could ask for a `list[str]` directly instead of allowing `str | list[str]`

requirements-common.txt
```diff
@@ -12,6 +12,7 @@ google-auth-oauthlib==1.2.1
 icalendar==5.0.13
 jellyfish==1.0.4                    # String Matching
 Jinja2==3.1.4                       # template engine for html files
+pandas==2.2.2
```
armanddidierjean, line 15: I'm not sure, but I think we could use python list instead of pandas dataframe

app/modules/cdr/endpoints_cdr.py
```diff
@@ -351,6 +355,197 @@ async def get_online_sellers(
     return await cruds_cdr.get_online_sellers(db)
 
 
+def construct_dataframe_from_users_purchases(
```
armanddidierjean, line 358: Move the util to `app/modules/cdr/utils_cdr.py`

app/utils/mail/mailworker.py
Rotheem: This is a convenient system which allow to not have inconsistency such as using a list of email when trying to send a personal token, moreover I didn't want to touch the old system

requirements-common.txt
Rotheem: This allows to create a ready-to-use excel, added to the fact that it was a way I knew and it took me far less time than finding a new one

app/modules/cdr/endpoints_cdr.py
Rotheem: The file creation is done by pandas it does not seem to be compatible

app/modules/cdr/endpoints_cdr.py
Rotheem: Hum, you might need to look at the endpoint_users.py file, there was quite a few missed, anyway it is fixed for this one

app/modules/cdr/endpoints_cdr.py
Rotheem: Fixed

app/modules/cdr/endpoints_cdr.py
Rotheem: Done

PR #563, opened 2024-09-04
User fusion

Rotheem:
### Description

### Checklist

- [ ] All tests passing


armanddidierjean (changes requested):
Could we add tests and documentation?

maximeroucher (approved):

Malem38 (approved):

app/core/users/endpoints_users.py
```diff
@@ -826,6 +826,39 @@ async def update_current_user(
...
+    if user_kept is None:
+        raise HTTPException(status_code=404, detail="User kept not found")
+
+    user_deleted = await cruds_users.get_user_by_email(
+        db=db,
+        email=user_fusion.user_deleted_email,
+    )
+    if user_deleted is None:
+        raise HTTPException(status_code=404, detail="User deleted not found")
+
```
armanddidierjean, line 854: Could you log a warning with the security logger indicating that two users (with email and user id) were merged?

app/core/users/endpoints_users.py
```diff
@@ -826,6 +826,39 @@ async def update_current_user(
     await cruds_users.update_user(db=db, user_id=user.id, user_update=user_update)
 
 
+@router.patch(
```
armanddidierjean, line 829: I think a POST may be more adapted as this start an action, and can not be called twice

app/core/users/endpoints_users.py
```diff
@@ -826,6 +826,39 @@ async def update_current_user(
...
+    "/users/fusion",
```
armanddidierjean, line 830:
```suggestion
    "/users/merge",
```

tests/test_user_fusion.py
```diff
@@ -0,0 +1,84 @@
...
+token_admin_user: str
+token_student_user: str
+
+
+FABRISTPP_EMAIL_1 = "fabristpp.eclair1@etu.ec-lyon.fr"
+FABRISTPP_EMAIL_2 = "fabristpp.eclair3@ecl21.ec-lyon.fr"
+
+
+@pytest_asyncio.fixture(scope="module", autouse=True)
+async def init_objects() -> None:
```
armanddidierjean, line 24: Could you add a few objects involving relationships? It would be great to have:
– an object with a relationship that is unique but not a primary key
– an object with a relationship that is neither unique nor a primary key

app/core/users/cruds_users.py
```diff
@@ -274,3 +275,122 @@ async def update_user_password_by_id(
...
+        We then update the row and check if the new row already exists:
+            a. If it doesn't, we update the row
+            b. If it does, we delete the row
+    """
+    foreign_keys: set[ForeignKey] = get_referencing_foreign_keys(
+        models_core.CoreUser.__table__,
+    )
+    user_id_fks: list[ForeignKey] = [
+        fk for fk in foreign_keys if fk.column == models_core.CoreUser.__table__.c.id
+    ]
```
armanddidierjean, line 311: Could you add a comment explaining the purpose of this comprehension? 
I believe the idea is to keep all relationships related of a CoreUser `id`, and remove other relationships that are related to another column of CoreUser

app/core/users/cruds_users.py
```diff
@@ -274,3 +275,122 @@ async def update_user_password_by_id(
...
+    for fk in user_id_fks:
+        is_in = False
+        for pk in fk.parent.table.primary_key.columns:  # "in" doesn't work with columns
+            if fk.parent == pk:
+                is_in = True
+                break
```
armanddidierjean, line 317: I would suggest rewording this comment, to make it clear we talk of the conditionnal `if x in list` instead of `for x in list`, 
```suggestion

        # We can not use `if fk.parent in fk.parent.table.primary_key.columns` because `in` doesn't work with columns
        for pk in fk.parent.table.primary_key.columns:
            if fk.parent == pk:
                is_in = True
                break
```

armanddidierjean (commented):
Impressive feature, thanks! I think we should speak of `merging user` (eventually `user fusion`) instead of `fusion user`

PR #569, opened 2024-09-08
Feat: clean and bump version

Rotheem:
### Description

### Checklist

- [ ] All tests passing

### DIFF
--- a/app/core/users/endpoints_users.py
+++ b/app/core/users/endpoints_users.py
@@ -826,18 +826,23 @@ async def update_current_user(
     await cruds_users.update_user(db=db, user_id=user.id, user_update=user_update)
 
 
-@router.patch(
-    "/users/fusion",
+@router.post(
+    "/users/merge",
     status_code=204,
 )
-async def fusion_users(
+async def merge_users(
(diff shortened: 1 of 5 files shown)


armanddidierjean (changes requested):
We need to address all user fusion and cdr recap issues before changing the version

maximeroucher (approved):

app/core/users/endpoints_users.py
```diff
@@ -857,6 +862,16 @@ async def fusion_users(
         user_kept_id=user_kept.id,
         user_deleted_id=user_deleted.id,
     )
+    background_tasks.add_task(
+        send_email,
+        recipient=[user_kept.email, user_deleted.email],
+        subject="MyECL - Account fusion",
```
maximeroucher, line 868:
```suggestion
        subject="MyECL - Fusion de compte",
```

maximeroucher: The whole mail content is in french, so why don't we go for the title as well

guillaumetavernier (approved):

Malem38 (approved):

PR #574, opened 2024-09-10
Add ticket tags lists

Malem38:
### Description

### Checklist

- [ ] All tests passing


app/modules/cdr/cruds_cdr.py
```diff
@@ -1016,3 +1016,39 @@ async def delete_product_generated_tickets(db: AsyncSession, ticket_generator_id
...
+    tag: str,
+) -> Sequence[models_cdr.Ticket]:
+    result = await db.execute(
+        select(models_cdr.Ticket)
+        .where(
+            models_cdr.Ticket.generator_id == generator_id,
+            models_cdr.Ticket.tags.contains(tag.lower()),
+        )
+        .options(
+            noload(models_cdr.Ticket.product_variant),
```
armanddidierjean, line 1033: Could you add a comment explaining why we need `noload`?

app/modules/cdr/endpoints_cdr.py
```diff
@@ -2760,14 +2760,109 @@ async def scan_ticket(
             db=db,
             ticket_id=ticket.id,
             scan=ticket.scan_left - 1,
-            tags=ticket.tags + "," + ticket_data.tag.strip(),
+            tags=ticket.tags + "," + ticket_data.tag.strip()
```
armanddidierjean, line 2763: If you need to strip, you could do it with a validator in the model, like the lowercase step

app/modules/cdr/schemas_cdr.py
```diff
@@ -296,7 +296,7 @@ class Ticket(BaseModel):
 
 
 class TicketScan(BaseModel):
-    tag: str
+    tag: Annotated[str, StringConstraints(to_lower=True)]
```
armanddidierjean, line 299: I think comma should be forbidden in tags, to prevent parsing issues

app/modules/cdr/endpoints_cdr.py
```diff
@@ -2760,14 +2760,109 @@ async def scan_ticket(
...
+            detail="This Ticket generator is not related to this product.",
+        )
+
+    tickets = await cruds_cdr.get_tickets_by_generator(db=db, generator_id=generator_id)
+
+    tags = set()
+    for ticket in tickets:
+        for tag in ticket.tags.split(","):
+            tags.add(tag)
+    return list(tags)
```
armanddidierjean, line 2863: I'm really not sure, but couldn't we simplify this?
```suggestion
    return set(ticket.tags.split(","))
```

armanddidierjean: If there is a type issue, maybe `return list(set(ticket.tags.split(",")))`

armanddidierjean (changes requested):
Could you add a comment in the model explaining `tags` is a comma separated string?

app/modules/cdr/endpoints_cdr.py
Malem38: It won't work

Rotheem (approved):

PR #578, opened 2024-09-21
Support multiple auth tentative for GoogleAPI

armanddidierjean:
### Description

Google API authentication is an oauth flow. To trust the callback called by google, we ensure that the provided `states` is the same as the one we got during the first step of the authentication.
As we only stored one `state` in a CoreData object in the database, the following situation could happen:
- we start the auth process, and get a state `state1`
- we start another auth process, and get a state `state2`, which replace the previous state in the db
- the first process finish, but then we have `state1 != state2`, we raise an error

We can now store multiple states in a dedicated table of the database. The states will be purged once an authentication tentative succeed.

I thought of using a CoreData containing a list of state instead of a complete new table, but this would probably be impracticable. Two workers may try to get, then update the coredata at the same time, to add a state each, resulting in loosing one of them.

### Checklist

- [ ] All tests passing


app/core/google_api/endpoints_google_api.py
```diff
@@ -0,0 +1,37 @@
...
+            request=request,
+        )
+
+    except Exception:
+        hyperion_error_logger.exception(
+            "Google API authentication callback error",
+        )
+        return "An error occurred during the Google API authentication callback"
+    else:
+        return "Ok"
```
maximeroucher, line 37: You can remove the else statement

app/core/google_api/models_google_api.py
maximeroucher, line 1: Why do we need model, isn't the coredata enough ?

maximeroucher (approved):

Foukki (approved):

PR #580, opened 2024-09-25
Remove deprecated base image and use uv

julien4215:


start.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+set -e
+
+python3 init.py
+exec fastapi run app/main.py "$@"
```
Rotheem, line 5: Use environment variables instead of command arguments
```suggestion
exec fastapi run app/main.py --workers ${NB_WORKERS:1}
```

docker-compose.yaml
```diff
@@ -32,6 +32,7 @@ services:
     image: hyperion-app
     container_name: hyperion-app
     restart: unless-stopped
+    command: ["--port", "80"]
```
Rotheem, line 35: There is no need to change port through the docker-compose
```suggestion
```

Rotheem (approved):

armanddidierjean (changes requested):

Foukki (approved):

PR #590, opened 2024-09-29
Ignore "not enough segment" error with JWT

Rotheem:
### Description

### Checklist

- [ ] All tests passing

### DIFF
--- a/app/utils/auth/auth_utils.py
+++ b/app/utils/auth/auth_utils.py
@@ -2,7 +2,7 @@ import logging
 
 import jwt
 from fastapi import HTTPException, status
-from jwt.exceptions import ExpiredSignatureError, InvalidTokenError
+from jwt.exceptions import DecodeError, ExpiredSignatureError, InvalidTokenError
 from pydantic import ValidationError
 from sqlalchemy.ext.asyncio import AsyncSession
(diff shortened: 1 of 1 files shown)


app/utils/auth/auth_utils.py
```diff
@@ -49,6 +49,14 @@ def get_token_data(
...
             detail="Token has expired",
         ) from error
+    except DecodeError as error:
+        hyperion_access_logger.exception(
+            f"Get_token_data: Failed to decode a token ({request_id})",
+        )
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="Could not validate credentials",
+        ) from error
```
maximeroucher, line 59: Why is the from error required here ?

Rotheem: It is currently recommended by Python's Code of Conduct when using try except

maximeroucher (approved):

Malem38 (approved):

julien4215 (changes requested):
I think we should remove all the `from error` in this file. I don't understand why we are using them here.

julien4215 (approved):

PR #596, opened 2024-10-03
fix: Raid issues

maximeroucher:
- Fix team number

### DIFF
--- a/app/modules/raid/cruds_raid.py
+++ b/app/modules/raid/cruds_raid.py
@@ -1,7 +1,7 @@
 from collections.abc import Sequence
 from datetime import UTC, datetime
 
-from sqlalchemy import delete, func, or_, select, update
+from sqlalchemy import and_, delete, func, or_, select, update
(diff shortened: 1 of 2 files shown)


Malem38 (approved):

armanddidierjean (commented):
Could you please add an unit test?

maximeroucher:
What do you want to test ?

armanddidierjean:
I think `set_team_number` is never called by the test, it is not covered

tests/test_raid.py
```diff
@@ -280,6 +366,17 @@ def test_update_team(client: TestClient):
     assert response.status_code == 204
 
 
+def set_team_number(client: TestClient):
```
armanddidierjean, line 369:
```suggestion
def test_set_team_number(client: TestClient):
```

armanddidierjean (changes requested):

armanddidierjean (approved):

PR #613, opened 2024-10-29
Send notification content through Firebase and add an Arq scheduler

foucblg:


app/core/endpoints_core.py
armanddidierjean, line 1: We need to remove these test changes

app/types/scheduler.py
```diff
@@ -0,0 +1,81 @@
...
+    scheduler = Scheduler(settings)
+    if scheduler.settings.host != "":
+        await scheduler.async_init()
+    return scheduler
+
+
+class Scheduler:
+    """Disappears sometimes for no reason"""
+
+    def __init__(self, redis_settings):
```
armanddidierjean, line 22: Could you type `redis_settings`?

app/types/scheduler.py
```diff
@@ -0,0 +1,81 @@
...
+    async def async_init(self):
+        if self.settings.host != "":
+            self.redis_pool = await create_pool(self.settings)
+            scheduler_logger.debug(f"Pool in init {self.redis_pool}")
+
+    async def queue_job_time_defer(
+        self,
+        job_function: Callable[..., Coroutine[Any, Any, Any]],
+        job_id: str,
+        defer_seconds: float,
```
armanddidierjean, line 34:
```suggestion
        defer_seconds: int,
```

armanddidierjean: I think you could pass a timedelta directly. Other modules will probably construct a timedelta to deduce the number of seconds.

```suggestion
        defer_by: timedelta,
```

app/types/scheduler.py
```diff
@@ -0,0 +1,81 @@
...
+                "run_task",
+                job_function=job_function,
+                _job_id=job_id,
+                _defer_until=defer_date,
+                **kwargs,
+            )
+            scheduler_logger.debug(f"Job {job_id} queued {job}")
+            return job
+
+    async def cancel_job(self, job_id):
```
armanddidierjean, line 74:
```suggestion
    async def cancel_job(self, job_id: str):
```

app/utils/communication/future_notifications.py
```diff
@@ -0,0 +1,68 @@
...
+from app.core.notification.notification_types import CustomTopic
+from app.core.notification.schemas_notification import Message
+from app.types.scheduler import Scheduler
+from app.utils.communication.notifications import NotificationManager, NotificationTool
+
+
+class FutureNotificationTool(NotificationTool):
+    def __init__(
+        self,
+        background_tasks: BackgroundTasks,
```
armanddidierjean, line 15: The `FutureNotificationTool` don't need to use a `BackgroundTasks`. Deferred tasks will be run in a dedicated worker

app/worker.py
```diff
@@ -0,0 +1,26 @@
...
+
+scheduler_logger = logging.getLogger("scheduler")
+
+
+async def run_task(ctx, job_function, **kwargs):
+    scheduler_logger.debug(f"Job function consumed {job_function}")
+    return await job_function(**kwargs)
+
+
+settings = get_settings()  # Ce fichier ne doit être lancé que par la prod
```
armanddidierjean, line 15:
```suggestion
# This bash script will only be run on production. We can use `get_settings()` directly which will get production settings
settings = get_settings()
```

Daihecyy (approved):

armanddidierjean (changes requested):
I think we need to subscribe new firebase tokens to the user's registered topics in `register_firebase_device`, and maybe to unsubscribe tokens in `unregister_firebase_device`

app/core/endpoints_core.py
```diff
@@ -28,7 +30,10 @@
     response_model=schemas_core.CoreInformation,
     status_code=200,
 )
-async def read_information(settings: Settings = Depends(get_settings)):
+async def read_information(
+    settings: Settings = Depends(get_settings),
+    scheduler: Scheduler = Depends(get_scheduler),
+):
```
armanddidierjean, line 36: Please revert

app/dependencies.py
```diff
@@ -164,6 +164,13 @@ def get_redis_client(
     return redis_client
 
 
+def get_scheduler(settings: Settings = Depends(get_settings)) -> Scheduler:
+    global scheduler
+    if scheduler is None:
+        scheduler = Scheduler() if settings.REDIS_HOST != "" else DummyScheduler()
```
armanddidierjean, line 170: Instead of using a `DummyScheduler`, we could add a check in `Scheduler` to verify if the redis host is defined. If this is not the case we can maybe drop tasks

app/modules/advert/endpoints_advert.py
```diff
@@ -266,12 +266,9 @@ async def create_advert(
     except ValueError as error:
         raise HTTPException(status_code=400, detail=str(error))
     message = Message(
-        context=f"advert-new-{id}",
-        is_visible=True,
         title=f"📣 Annonce - {result.title}",
         content=result.content,
-        # The notification will expire in 3 days
-        expire_on=datetime.now(UTC) + timedelta(days=3),
+        action_module="advert",
```
armanddidierjean, line 271: What would you think of using the module root?
```suggestion
        action_module=module.root,
```

app/modules/ph/endpoints_ph.py
```diff
@@ -124,21 +127,25 @@ async def create_paper(
...
-                ),
-                # The notification will expire in 10 days
-                expire_on=now + timedelta(days=10),
-            )
-            await notification_tool.send_notification_to_topic(
-                custom_topic=CustomTopic(topic=Topic.ph),
-                message=message,
+                action_module="ph",
             )
+            if paper_db.release_date == now.date():
```
armanddidierjean, line 134: Could you defer the notification sending even if the delivery date is the current day?

app/types/scheduler.py
```diff
@@ -0,0 +1,229 @@
...
+
+async def run_task(
+    ctx,
+    job_function: Callable[..., Coroutine[Any, Any, Any]],
+    _get_db: Callable[[], AsyncGenerator[AsyncSession, None]],
+    **kwargs,
+):
+    scheduler_logger.debug(f"Job function consumed {job_function}")
+
+    if "db" in kwargs:
```
armanddidierjean, line 27: TODO

app/types/scheduler.py
```diff
@@ -0,0 +1,229 @@
...
+        """
+
+        class ArqWorkerSettings(WorkerSettingsBase):
+            functions = [run_task]
+            allow_abort_jobs = True
+            keep_result_forever = True
+            redis_settings = RedisSettings(
+                host=redis_host,
+                port=redis_port,
+                password=redis_password if redis_password is not None else "",
```
armanddidierjean, line 75: Does this works if we pass `""` as password?

armanddidierjean (changes requested):
We should add a test that tries to defer a method using the scheduler

app/modules/ph/endpoints_ph.py
foucblg: Why ?

app/types/scheduler.py
Daihecyy: ?

app/dependencies.py
Daihecyy: Yes but we would still need to call the methods with no implementation when sending notifications thus the dummy Scheduler.
Thonyk said it was ok

app/types/scheduler.py
Daihecyy: Empty password is allowed depending if redis has no password set up

armanddidierjean: Brilliant, thanks
```suggestion
                password=redis_password or "",
```

app/modules/ph/endpoints_ph.py
armanddidierjean: This would simplify the code as the module would not have to check the current date and decide if the notification needs to be planned or send now.
I think this should not have a real impact on performance
 
I believe using the scheduler to send notifications in the future is specific of our Firebase implementation. The module is not forced to know how we implemented this feature to send notifications. For example with the previous implementation the module did not have to check when the notification was needed to be send

foucblg: Not agree

app/dependencies.py
armanddidierjean: This indeed works, but we never created a *dummy* class like this one in the rest of the code. We usually check if Redis/Firebase is configured, and if this is not the case, we just return without executing the rest of the method.

I would think that using this solution would be more coherent with the rest of our code base. Moreover, it only adds a few lines of code instead of defining a complete new class `DummyScheduler`

app/modules/loan/endpoints_loan.py
```diff
@@ -872,18 +872,7 @@ async def return_loan(
...
-        is_visible=False,
-        title="",
-        content="",
-        expire_on=datetime.now(UTC) + timedelta(days=3),
-    )
-    await notification_tool.send_notification_to_user(
-        user_id=loan.borrower_id,
-        message=message,
-    )
+    await scheduler.cancel_job(f"loan_end_{loan.id}")
```
armanddidierjean, line 875: As the job was created by the `notification_manager` which mask the usage of the scheduler, you could create a `notification_manager.cancel_notification(notification_id: str)`

armanddidierjean: I think the method that send a future notification could always cancel existing notifications with the same id. I don't think there will be a case where we want to send two notifications with the same id

app/types/scheduler.py
```diff
@@ -0,0 +1,229 @@
...
+
+scheduler_logger = logging.getLogger("scheduler")
+
+
+async def run_task(
+    ctx,
+    job_function: Callable[..., Coroutine[Any, Any, Any]],
+    _get_db: Callable[[], AsyncGenerator[AsyncSession, None]],
+    **kwargs,
+):
```
armanddidierjean, line 24: What happens if this method raises an error? 

If the exception is caught by Arq (I hope and believe this should be the case), we should wrap the body of this method in a try/except statement. If an exception occurs, we want to log the exception trace, maybe the job id, and reraise the exception to let Arq know the job failed.

app/types/scheduler.py
```diff
@@ -0,0 +1,229 @@
...
+            "run_task",
+            job_function=job_function,
+            _job_id=job_id,
+            _defer_by=timedelta(seconds=defer_seconds),
+            _get_db=self._get_db,
+            **kwargs,
+        )
+        scheduler_logger.debug(f"Job {job_id} queued {job}")
+
+    async def queue_job_defer_to(
```
armanddidierjean, line 121: Is there a real interest in keeping both `queue_job_time_defer` and `queue_job_defer_to`? 
I think we are likely to calculate `defer_seconds` using a `timedelta` and `datetime`, and thus could easily use `queue_job_defer_to` instead.

It seems Arq will calculate a `datetime` in both cases, so there shouldn't be a performance gain from using a method or the other.

I would think only having one method will reduce code complexity and limit potentials mistakes

app/dependencies.py
Daihecyy: This does work but is however not coherent with using dependency injection. Doing nothing seems more of a get around than injecting a custom mock implementation.
If consistency is an issue, it might be worthwhile considering to change the rest and not this change.

app/core/endpoints_core.py
Rotheem: This comment doesn't mean anything

app/core/endpoints_core.py
```diff
@@ -28,7 +30,10 @@
     response_model=schemas_core.CoreInformation,
     status_code=200,
 )
-async def read_information(settings: Settings = Depends(get_settings)):
+async def read_information(
+    settings: Settings = Depends(get_settings),
+    scheduler: Scheduler = Depends(get_scheduler),
```
Daihecyy, line 35:
```suggestion
```

app/core/endpoints_core.py
Daihecyy: has been fixed

armanddidierjean (approved):
We will improve the remaining points in the future
https://github.com/aeecleclair/Hyperion/issues/647

PR #620, opened 2024-11-25
Ignore RedisConnectionError and exclude TypeChecking blocks from coverage

armanddidierjean:


Daihecyy (commented):
Consider changing `xfail` to [`skipif`](https://docs.pytest.org/en/stable/how-to/skipping.html#id1) since the tests seem to be expected to run in Github actions.
Maybe create a shareable skipif decorator in `test_commons` ?

Daihecyy (changes requested):
Consider changing `xfail` to [`skipif`](https://docs.pytest.org/en/stable/how-to/skipping.html#id1) since the tests seem to be expected to run in Github actions.
Maybe create a shareable skipif decorator in `test_commons` ?

Daihecyy:
> > Consider changing `xfail` to [`skipif`](https://docs.pytest.org/en/stable/how-to/skipping.html#id1) since the tests seem to be expected to run in Github actions. Maybe create a shareable skipif decorator in `test_commons` ?
> 
> I'm not sure how I can decide if the test should be run as settings are not loaded when the file is imported. I think that failed test should raise an error in github action

But the tests passing would be considered "unexpedly passing" `XPASS` which is not really what we want. 
We could do a similar thing as with `override_get_db` in `tests/commons` and return None if the redis client is not configured.
We could then `skipif` with the condition the redis client being None.
Such an approach would be a lot cleaner as we would not allow our tests to fail.

Daihecyy:
I must say that the way `redis_client` is handled in `tests/commons` is really weird to me. 
Such a type `redis.Redis | None | bool` seems to indicate something is wrong.

PR #625, opened 2024-12-02
Use server url defined in dotenv

armanddidierjean:
instead of matrix.org


Rotheem (approved):

maximeroucher (approved):

Daihecyy:
I think having only one url will reduce code complexity

PR #628, opened 2024-12-04
Flappybird security

Riad-Attou:
Add security for score submission with key verification in flappybird

Previously, intercepting requests allowed users to modify scores. 
Now, an additional security layer with a key mechanism ensures 
scores cannot be modified with during transmission.

This commit goes hand-in-hand with the related changes in Titan.

### DIFF
--- a/app/modules/flappybird/models_flappybird.py
+++ b/app/modules/flappybird/models_flappybird.py
@@ -14,6 +14,7 @@ class FlappyBirdScore(Base):
     user_id: Mapped[str] = mapped_column(ForeignKey("core_user.id"))
     user: Mapped[CoreUser] = relationship("CoreUser", init=False)
     value: Mapped[int]
+    key: Mapped[int]
     creation_time: Mapped[datetime]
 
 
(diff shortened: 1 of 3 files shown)


Daihecyy:
See the section on linting and formatting to allow your code to pass : https://github.com/aeecleclair/Hyperion?tab=readme-ov-file#linting-and-formating 
Otherwise, great job !

app/modules/flappybird/models_flappybird.py
```diff
@@ -14,6 +14,7 @@ class FlappyBirdScore(Base):
     user_id: Mapped[str] = mapped_column(ForeignKey("core_user.id"))
     user: Mapped[CoreUser] = relationship("CoreUser", init=False)
     value: Mapped[int]
+    key: Mapped[int]
```
julien4215, line 17: Why are we storing the key in the database ?

Riad-Attou: > Why are we storing the key in the database ?

That was a mistake, I think it's fixed now.

PR #633, opened 2024-12-18
Bump version 4.0.0

Marc-Andrieu:


armanddidierjean (changes requested):
I think we can bump a minor or a major

PR #641, opened 2024-12-22
Schools

Rotheem:
### Description

### Checklist

- [x] All tests passing


app/core/schools/endpoints_schools.py
```diff
@@ -0,0 +1,185 @@
...
+            detail="A school with the name {school.name} already exist",
+        )
+
+    try:
+        db_school = models_core.CoreSchool(
+            id=str(uuid.uuid4()),
+            name=school.name,
+            email_regex=school.email_regex,
+        )
+        new_school = await cruds_schools.create_school(school=db_school, db=db)
```
Daihecyy, line 99: By putting the commit in `create_school` and in `update_user`, we cannot rollback the entire endpoint.
Dont we risk having updated only half of the users ? (then again we could just rerun the endpoint)

app/core/schools/schools_type.py
```diff
@@ -0,0 +1,16 @@
+from enum import Enum
+
+
+class SchoolType(str, Enum):
+    """
+    In Hyperion, each user must have a school. Belonging to a school gives access to a set of specific endpoints.
```
Daihecyy, line 6: The name SchoolType and what exactly is does is confusing, could you add more comments ?
(for ex what about students that are neither `no_school` and `centrale_lyon`)

tests/test_schools.py
Daihecyy, line 1: School deletion not tested

app/core/models_core.py
```diff
@@ -118,6 +125,21 @@ class CoreGroup(Base):
     )
 
 
+class CoreSchool(Base):
+    __tablename__ = "core_school"
+
+    id: Mapped[str] = mapped_column(String, primary_key=True, index=True)
```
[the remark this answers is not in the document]
Rotheem: For now UUID has only caused issues and no benefits. I still consider their use more troublesome than strings

app/core/schools/cruds_schools.py
```diff
@@ -0,0 +1,85 @@
+"""File defining the functions called by the endpoints, making queries to the table using the models"""
```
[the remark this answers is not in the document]
Rotheem: I copied groups' file

app/core/groups/groups_type.py
```diff
@@ -57,7 +58,7 @@ def get_ecl_account_types() -> list[AccountType]:
     ]
 
 
-def get_account_types_except_external() -> list[AccountType]:
+def get_account_types_except_externals() -> list[AccountType]:
```
armanddidierjean, line 61: Should we return `other_school_student`?

app/core/models_core.py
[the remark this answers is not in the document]
armanddidierjean: Please still use a UUID as this is currently considered the standard for new models

app/core/schools/endpoints_schools.py
```diff
@@ -0,0 +1,215 @@
...
+
+@router.get(
+    "/schools/{school_id}",
+    response_model=schemas_core.CoreSchool,
+    status_code=200,
+)
+async def read_school(
+    school_id: str,
+    db: AsyncSession = Depends(get_db),
+    user=Depends(is_user_in(GroupType.admin)),
```
armanddidierjean, line 54:
```suggestion
    user: schemas_core.CoreUser = Depends(is_user_in(GroupType.admin)),
```

app/core/schools/endpoints_schools.py
```diff
@@ -0,0 +1,215 @@
...
+    response_model=schemas_core.CoreSchoolSimple,
+    status_code=201,
+)
+async def create_school(
+    school: schemas_core.CoreSchoolBase,
+    db: AsyncSession = Depends(get_db),
+    user: schemas_core.CoreUser = Depends(is_user_in(GroupType.admin)),
+):
+    """
+    Create a new school.
```
armanddidierjean, line 79: We need to indicate that external users may become members of this new school.

app/core/schools/endpoints_schools.py
```diff
@@ -0,0 +1,215 @@
...
+                await cruds_users.update_user(
+                    db,
+                    db_user.id,
+                    schemas_core.CoreUserUpdateAdmin(
+                        school_id=db_school.id,
+                        account_type=AccountType.other_school_student,
+                    ),
+                )
+        await db.commit()
+    except ValueError as error:
```
armanddidierjean, line 114: In what situation could we have a `ValueError`?

app/core/schools/endpoints_schools.py
```diff
@@ -0,0 +1,215 @@
...
+        await db.rollback()
+        raise HTTPException(status_code=422, detail=str(error))
+    except IntegrityError:
+        await db.rollback()
+        raise HTTPException(
+            status_code=400,
+            detail="School creation failed due to database integrity error",
+        )
```
armanddidierjean, line 122: I think letting FastAPI return a 500 may be pertinent as this should never happen. This would allow us to log the entire trace.

app/core/schools/endpoints_schools.py
```diff
@@ -0,0 +1,215 @@
...
+        for db_user in users:
+            if re.match(db_school.email_regex, db_user.email):
+                await cruds_users.update_user(
+                    db,
+                    db_user.id,
+                    schemas_core.CoreUserUpdateAdmin(
+                        school_id=db_school.id,
+                        account_type=AccountType.other_school_student,
+                    ),
+                )
```
armanddidierjean, line 112: This is clever, be I think it would be great to have a similar comportment on update. This is strange to modify users on school creation but on school modification.
Doing so is also a bit dangerous. If an admin makes a mistake when setting the regex, they may not be able to revert the induced changes.

app/core/models_core.py
```diff
@@ -118,6 +125,21 @@ class CoreGroup(Base):
...
+    name: Mapped[str] = mapped_column(String, index=True, nullable=False, unique=True)
+    email_regex: Mapped[str] = mapped_column(String, nullable=False)
+
+    students: Mapped[list["CoreUser"]] = relationship(
+        "CoreUser",
+        back_populates="school",
+        lazy="selectin",
+        default_factory=list,
+    )
```
armanddidierjean, line 140: I don't think this is really useful to index `name`. I think this would only be used to check another school does not exist with the same name, which should not happen often.

I would say we don't need to init the `students` list, but as you prefer

```suggestion
    id: Mapped[PrimaryKey]
    name: Mapped[str] = mapped_column(unique=True)
    email_regex: Mapped[str]

    students: Mapped[list["CoreUser"]] = relationship(
        "CoreUser",
        back_populates="school",
        lazy="selectin",
        init=False,
    )
```

app/core/schools/endpoints_schools.py
```diff
@@ -0,0 +1,215 @@
...
+            detail="SchoolTypes schools can not be deleted",
+        )
+
+    school = await cruds_schools.get_school_by_id(db=db, school_id=school_id)
+    if school is None:
+        raise HTTPException(status_code=404, detail="School not found")
+
+    users = await cruds_users.get_users(db=db, schools_ids=[school_id])
+    for db_user in users:
+        await cruds_users.update_user(
```
armanddidierjean, line 198: Technically users emails may match another school

armanddidierjean (changes requested):

app/core/models_core.py
```diff
@@ -118,6 +125,21 @@ class CoreGroup(Base):
...
 
 
+class CoreSchool(Base):
+    __tablename__ = "core_school"
+
+    id: Mapped[str] = mapped_column(String, primary_key=True, index=True)
+    name: Mapped[str] = mapped_column(String, index=True, nullable=False, unique=True)
+    email_regex: Mapped[str] = mapped_column(String, nullable=False)
+
+    students: Mapped[list["CoreUser"]] = relationship(
```
armanddidierjean, line 135: May not be useful

app/core/schemas_core.py
```diff
@@ -19,6 +19,44 @@ class CoreInformation(BaseModel):
...
+    description: str | None = None
+
+    _normalize_name = field_validator("name")(validators.trailing_spaces_remover)
+
+
+class CoreGroupSimple(CoreGroupBase):
+    """Simplified schema for group's model, used when getting all groups"""
+
+    id: str
+    model_config = ConfigDict(from_attributes=True)
```
armanddidierjean, line 35: Is this needed? Could you add a comment explaining why?

armanddidierjean (changes requested):

app/core/schools/endpoints_schools.py
```diff
@@ -0,0 +1,232 @@
...
+    Delete school from database.
+    This will remove the school from all users but won't delete any user.
+
+    `SchoolTypes` schools can not be deleted.
+
+    **This endpoint is only usable by administrators**
+    """
+
+    if school_id in (item.value for item in SchoolType):
+        raise HTTPException(
```
armanddidierjean, line 205: Plse test

app/core/schools/endpoints_schools.py
```diff
@@ -0,0 +1,232 @@
...
+            detail="SchoolTypes schools can not be deleted",
+        )
+
+    school = await cruds_schools.get_school_by_id(db=db, school_id=school_id)
+    if school is None:
+        raise HTTPException(status_code=404, detail="School not found")
+
+    users = await cruds_users.get_users(db=db, schools_ids=[school_id])
+    for db_user in users:
+        await cruds_users.update_user(
```
armanddidierjean, line 216: Please test

app/core/users/cruds_users.py
```diff
@@ -72,6 +75,15 @@ async def get_users(
                     )
                     for group_id in excluded_groups
                 ],
+                or_(
```
armanddidierjean, line 78: You can `append` the condition `or_()` if `schools_ids`

armanddidierjean (changes requested):

app/core/schools/endpoints_schools.py
```diff
@@ -0,0 +1,215 @@
...
+router = APIRouter(tags=["Schools"])
+
+
+@router.get(
+    "/schools/",
+    response_model=list[schemas_core.CoreSchool],
+    status_code=200,
+)
+async def read_schools(
+    db: AsyncSession = Depends(get_db),
```
Daihecyy, line 33: Consider putting administrator only if you put `schools/{school_id}` in administrator only since this one returns all the possible results of `schools/{school_id}` and thus should be at least of same privilege

Daihecyy (approved):

app/app.py
```diff
@@ -189,6 +190,32 @@ def initialize_groups(
...
+    hyperion_error_logger.info("Startup: Adding new groups to the database")
+    with Session(sync_engine) as db:
+        for school in SchoolType:
+            exists = initialization.get_school_by_id_sync(school_id=school.value, db=db)
+            # We don't want to recreate the groups if they already exist
+            if not exists:
+                db_school = models_core.CoreSchool(
+                    id=school.value,
+                    name=school.name,
+                    email_regex=".*",
```
armanddidierjean, line 208: We should probably set a regex

app/app.py
```diff
@@ -189,6 +190,32 @@ def initialize_groups(
...
+                )
+
+                try:
+                    initialization.create_school_sync(school=db_school, db=db)
+                except IntegrityError as error:
+                    hyperion_error_logger.fatal(
+                        f"Startup: Could not add group {db_school.name}<{db_school.id}> in the database: {error}",
```
armanddidierjean, line 215:
```suggestion
                        f"Startup: Could not add school {db_school.name}<{db_school.id}> in the database: {error}",
```

app/core/models_core.py
```diff
@@ -118,6 +125,14 @@ class CoreGroup(Base):
     )
 
 
+class CoreSchool(Base):
+    __tablename__ = "core_school"
+
+    id: Mapped[PrimaryKey]
+    name: Mapped[str] = mapped_column(String, unique=True)
+    email_regex: Mapped[str] = mapped_column(String)
```
armanddidierjean, line 133:
```suggestion
    name: Mapped[str] = mapped_column(unique=True)
    email_regex: Mapped[str]
```

armanddidierjean (approved):

PR #652, opened 2025-01-12
Add PlankaAuthClient

Marc-Andrieu:
Because Trello boards with several people aren't free anymore, BDE's president Larry asked last Tuesday for a Trello alternative for kanban boards, with possible MyECL integration. Here is the auth client class.

Goes hand in hand with [compose#137](../../compose/pull/137)

### DIFF
--- a/app/utils/auth/providers.py
+++ b/app/utils/auth/providers.py
@@ -418,12 +418,6 @@ class PlankaAuthClient(BaseAuthClient):
 
     @classmethod
     def get_userinfo(cls, user: models_core.CoreUser):
-        # Must match ^[a-zA-Z0-9]+((_|\.)?[a-zA-Z0-9])*$
-        username = unidecode.unidecode(
-            f"{user.firstname.strip()}.{user.name.strip()}",
-        ).replace(" ", "_")
-        username = re.sub(r"[^a-zA-Z0-9._]", "", username)
-
         return {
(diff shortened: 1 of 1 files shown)


app/utils/auth/providers.py
```diff
@@ -400,3 +400,31 @@ def get_userinfo(cls, user: models_core.CoreUser):
...
+
+class PlankaAuthClient(BaseAuthClient):
+    """
+    An auth client for Planka, a Trello alternative for kanban boards
+
+    Docs for OIDC integration:
+    https://docs.planka.cloud/docs/Configuration/OIDC/
+    """
+
+    allow_pkce_with_client_secret: bool = True  # required to exchange OIDC code
```
armanddidierjean, line 413: To be honest I'm not sure if I understood correctly the comment. What did you mean?
```suggestion
    allow_pkce_with_client_secret: bool = True
```

Marc-Andrieu: In practice, if this boolean is left to false, Planka throws an error right after the redirection from CalypSSO.
It says that (IIRC) the "exchange of the OIDC codes using PKCE" failed.
This message appears both in Planka's logs and right after the callback in the response to a failed request.

This led me to look at this boolean about PKCE.
When I set it to true, the authentication process worked as expected.

foucblg (approved):

PR #665, opened 2025-02-17
CMM

Marc-Andrieu:
[The PR for this branch was missing]


Marc-Andrieu:
Front-end counterpart: [Titan#461](../../Titan/pull/461)

Daihecyy (commented):
Removing page feature will lead to huge trafic (possibly thousands of memes sent each time when only the first 20 will likely be seen)

PR #672, opened 2025-02-24
Fix attempt: run "expected" workflows

Marc-Andrieu:
### Problem

* No PR can be merged, unless an admin makes use of his power to bypass the requirements.
* Because *Analyze (python)* and *CodeQL* are stuck as "pending checks", since:
  * They are expected to run
  * They don't run: the default config (overridden by this YAML) made CodeQL be triggered only on push main

### What we want

* We want those workflows to be triggered on PRs like *lint and format* and *test*.
* I open a PR so that this YAML can be reviewed.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,44 @@
+name: "CodeQL Advanced"
+
+on:
+  pull_request:
+    types: [opened, edited, ready_for_review, synchronize]
+  push:
+    branches:
+      - main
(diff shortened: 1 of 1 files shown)


Marc-Andrieu:
Update: "All checks have passed".
But in the first place: all checks have run!

maximeroucher (approved):
Since they are running, it seems to be working

Rotheem (approved):

PR #686, opened 2025-03-18
AMAP in cents

Flaim-X2:
Replace float with int.


armanddidierjean (changes requested):
Please decomment or remove commented code

armanddidierjean (approved):
LGTM, we should test it carefully with the front

Luhmos (approved):

PR #688, opened 2025-03-22
Grant BDE all the rights in Booking

Marc-Andrieu:
See [Titan#497](https://github.com/aeecleclair/Titan/pull/497)

Requested by BDE's secretary


Marc-Andrieu:
Decision we took at CoNum: keeping the creation/edition/deletion of rooms in the hands of the admin group

PR #697, opened 2025-05-01
Ruff 0.11.8

armanddidierjean:
### Description

Bump Ruff. Enable new rules


Marc-Andrieu (approved):
I approve because we need this PR and I've reviewed 43/46 files, but still don't get these two last changes :
* `app/core/core_endpoints/endpoints_core.py` : why remove the try-except block and the HTTP exception?
* why remove the parentheses after `@pytest.fixture`?

PR #700, opened 2025-05-02
Move versions declaration in pyproject.toml

Marc-Andrieu:
### Why

* Inspired by https://github.com/aeecleclair/Titan/pull/480
* Now both the version of itself and of the Titan client are declared in the TOML config file

### How

* I tried cleverly adding [`toml_file="pyproject.toml",`](https://docs.pydantic.dev/latest/concepts/pydantic_settings/#other-settings-source) in the `SettingsConfigDict` (line 49) so that the model would load fields from both config files (`.env` and `pyproject.toml`), but it seems Pydantic does not allow that at all...
* So I went using Python's *native* [`tomllib`](https://docs.python.org/3/library/tomllib.html) TOML parser : it the only solution that worked for me, and it's short.
  * (The `with` is out of the class because Pydantic started crying about the type of the bytes flow from reading the file)

### [`pyproject.toml`](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/)

* I just wanted to add
```toml
version = "4.3.3"
minimal-titan-version-code = 139
```
to the `[project]` table, but I realized it **didn't even exist**, so I ended up writing the bare minimum.
* Perhaps we should fill this file seriously...
  * it is powerful for automation and description (it could not hurt to have it filled),
  * and required to make a release on [PyPI](https://pypi.org/user/ECLAIR/) (we already release CalypSSO and the HA wrapper, why not Hyperion itself?).

### DIFF
--- a/app/core/utils/config.py
+++ b/app/core/utils/config.py
@@ -1,4 +1,6 @@
+import tomllib
 from functools import cached_property
+from pathlib import Path
 from typing import Any
 
 import jwt
@@ -14,6 +16,11 @@ from app.types.exceptions import (
     InvalidRSAKeyInDotenvError,
 )
 from app.utils.auth import providers
+
+with Path("pyproject.toml").open("rb") as pyproject_binary:
+    pyproject = tomllib.load(pyproject_binary)
+    HYPERION_VERSION: str = pyproject["project"]["version"]
+    MINIMAL_TITAN_VERSION_CODE: int = pyproject["project"]["minimal-titan-version-code"]
 
 
 class Settings(BaseSettings):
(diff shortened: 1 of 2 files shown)


app/core/utils/config.py
```diff
@@ -15,6 +17,11 @@
 )
 from app.utils.auth import providers
 
+with Path("pyproject.toml").open("rb") as pyproject_binary:
+    pyproject = tomllib.load(pyproject_binary)
+    HYPERION_VERSION: str = pyproject["project"]["version"]
+    MINIMAL_TITAN_VERSION_CODE: int = pyproject["project"]["minimal-titan-version-code"]
```
armanddidierjean, line 23: I suggest using a cached `computed_field`, like `RSA_PRIVATE_KEY` instead of a global variable that will be computed as soon as the file is imported

armanddidierjean (approved):
Really great!

app/core/utils/config.py
Marc-Andrieu: In this new commit, the TOML file is opened twice.

But perhaps it's better to be sure it will never be opened a 3rd time, rather than opening it at every use of the Settings class through the import of `config.py` ?

armanddidierjean: If you don't like the file to be open twice, you could create a private cached property which load the toml and then the two properties would use it to find the correct value:

```python
    @computed_field  # type: ignore[misc]
    @cached_property
    def PYPROJECT(cls) -> dict:
        with Path("pyproject.toml").open("rb") as pyproject_binary:
            return tomllib.load(pyproject_binary)

    @computed_field  # type: ignore[misc]
    @cached_property
    def HYPERION_VERSION(cls) -> str:
        return self.PYPROJECT["project"]["version"]

    @computed_field  # type: ignore[misc]
    @cached_property
    def MINIMAL_TITAN_VERSION_CODE(cls) -> str:
        return self.PYPROJECT["project"]["minimal-titan-version-code"]
```

Marc-Andrieu: But this will store in cache the content of the whole `pyproject.toml`... I guess opening the file twice but only on start is better. Unless there's some mixed solution?

PR #764, opened 2025-06-28
CdR: get the tickets generators of a product

Marc-Andrieu:
### Description

When editing a product on Siarnaq, we need to get the tickets generators associated to this product (now in the plural!), much like we already do for the custom questions to users (so-called "data fields").

### Checklist

- [x] Created an endpoint
- [x] Sort of created a schema for the response
- [x] Created a CRUD, fortunately using an existing model: the work is already done on the db side, this PR merely creates an interface to get a specific list

### DIFF
--- a/app/modules/cdr/cruds_cdr.py
+++ b/app/modules/cdr/cruds_cdr.py
@@ -797,6 +797,18 @@ async def get_ticket_by_secret(
     return result.scalars().first()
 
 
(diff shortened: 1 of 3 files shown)


Marc-Andrieu:
In fact, they were right, it is possible without that unnecessary custom hook that requires a new Hyperion endpoint.

PR #768, opened 2025-07-04
Acquire a lock before running initialization steps

armanddidierjean:
### Description

If multiple workers are detected, and a Redis client, Hyperion will raise an error.
Otherwise the server won't be able to:
 - run initialization steps
 - send messages over WebSocket reliably.


app/utils/initialization.py
```diff
@@ -269,3 +276,70 @@ def drop_db_sync(conn: Connection):
...
+
+        if unlock_key is not None:
+            # We set the unlock_key for other workers to resume operation
+            redis_client.set(unlock_key, "1")
+
+            # After 20 seconds, we assume that the job is done and we can release the lock
+            redis_client.expire(unlock_key, 60)
+
+        # After 20 seconds, we assume that the job is done and we can release the lock
+        redis_client.expire(key, 60)
```
Rotheem, line 326:
```suggestion
            # After 60 seconds we remove the key for both performance and reloading issues
            # we assume other jobs won't take more than 60 seconds and will check this key before expiration
            redis_client.expire(unlock_key, 60)

        # After 60 seconds we remove the key for both performance and reloading issues
        # we assume other jobs won't take more than 60 seconds and will check this key before expiration
        redis_client.expire(key, 60)
```

Rotheem (approved):

PR #805, opened 2025-08-28
Minor config fixes and enhancements

Marc-Andrieu:
While reconfiguring my local Hyperion, a couple difficulties arose...


.vscode/settings.json
```diff
@@ -25,9 +25,5 @@
     // on config.yaml and dotenv files
     "yaml": false,
     "dotenv": false
-  },
-  "files.associations": {
-    ".env": "dotenv",
-    ".env.template": "dotenv"
```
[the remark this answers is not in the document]
armanddidierjean: It is used to disable copilot in the dotenv

config.template.yaml
```diff
@@ -4,13 +4,40 @@
 # ACCESS_TOKEN_SECRET_KEY should contain a random string with enough entropy (at least 32 bytes long) to securely sign all access_tokens for OAuth and Openid connect
 ACCESS_TOKEN_SECRET_KEY: ""
 # RSA_PRIVATE_PEM_STRING should be a string containing the PEM certificate of a private RSA key. It will be used to sign id_tokens for Openid connect authentication
-# In the pem certificates newlines can be replaced by `\n`
-RSA_PRIVATE_PEM_STRING: ""
+# The example below was generated using a 2048-bit RSA key generator
+RSA_PRIVATE_PEM_STRING: |
```
armanddidierjean, line 8: I'm not fundamentally against providing a default PEM string for local development, but I am afraid some people may keep it for services exposed over internet

config.template.yaml
```diff
@@ -34,17 +61,34 @@ CLIENT_URL: "http://127.0.0.1:8000/"
...
-LOG_DEBUG_MESSAGES: true
+LOG_DEBUG_MESSAGES: True
 
 # Origins for the CORS middleware. `["http://localhost"]` can be used for development.
 # See https://fastapi.tiangolo.com/tutorial/cors/
 # It should begin with 'http://' or 'https:// and should never end with a '/'
-CORS_ORIGINS: ["http://localhost"]
+CORS_ORIGINS:
+  - http://localhost:3000
+  - http://127.0.0.1:3000/
```
armanddidierjean, line 90: We indicate a few lines earlier that these address should never end with a `/`. We should either change the requirement or the default value

.vscode/settings.json
```diff
@@ -25,9 +25,5 @@
     // on config.yaml and dotenv files
     "yaml": false,
     "dotenv": false
-  },
-  "files.associations": {
-    ".env": "dotenv",
```
armanddidierjean, line 30: We should ensure that Copilot remains disabled on the dotenv after this change

PR #824, opened 2025-09-05
Enable promo-based Piwigo groups

Marc-Andrieu:
Add the promo to the "groups" claim of OIDC's "API" scope in Piwigo's auth client


Rotheem (approved):

Rotheem (changes requested):
We might want to filter promotions being sent with the format 20xx as many users have wrongly formatted ones and each will create a useless group in piwigo

Rotheem (approved):

PR #835, opened 2025-09-15
added names to amap deliveries

cotanoine:
This PR adds names to deliveries from the Amap module


Marc-Andrieu:
Do only one migration

Marc-Andrieu:
Why do you add the name to the general `delivery` AND everyone's `order`s *independently*?
You can get the "order's name" from the delivery its related to by mapping the column.

PR #847, opened 2025-09-22
Fix  raid logic

maximeroucher:
### Description

Fixes team merge crashes because invite tokens still exists
Fixes raid price calculation

### Checklist

- [ ] All tests passing

### DIFF
--- a/tests/test_raid.py
+++ b/tests/test_raid.py
@@ -676,150 +676,150 @@ def test_delete_team(client: TestClient):
     assert response.status_code == 204
 
 
-def test_get_payment_url_student_price(client: TestClient, mocker):
-    mocker.patch(
-        "app.modules.raid.endpoints_raid.get_core_data",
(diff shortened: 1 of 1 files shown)


app/modules/raid/utils/utils_raid.py
```diff
@@ -574,3 +574,43 @@ async def generate_teams_pdf_util(
...
+    checkout_name = ""
+    if not participant:
+        raise HTTPException(status_code=404, detail="Participant not found.")
+
+    if not participant.payment:
+        if (
+            participant.situation
+            and participant.situation.split(" : ")[0] in ["centrale", "otherschool"]
+            and participant.student_card is not None
+            and participant.student_card.id is not None
```
armanddidierjean, line 600: Would checking `participant.student_card.id is not None` be enough?

armanddidierjean (approved):

PR #852, opened 2025-09-23
Deploy/single stage docker

warix8:
This pull request is part of a clarification process aimed at simplifying handovers and maintainability. 

Here, it is a clean-up of the Docker image build. 
The multi-stage build of the Docker image is not at all suitable in our case, and was used to create a kind of ‘cache’, which can in fact be easily resolved as mentioned in the FastAPI documentation as a Docker ‘trick’. The idea is to use --upgrade on pip install to take advantage of layer caching. [Reference FastAPI](https://fastapi.tiangolo.com/deployment/docker/?h=docke#docker-cache)
In addition, using a slim image means that only the strictly necessary tools are included.

A multi-stage build image is relevant in the case of a compilation where the code is subsequently executed in a lighter environment, which then contains fewer vulnerabilities and is lighter.
The publish.yml has therefore been restored to its original version.

### DIFF
--- a/.github/workflows/alpha-publish.yml
+++ b/.github/workflows/alpha-publish.yml
@@ -12,15 +12,15 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
         with:
           ref: "main"
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3
+        uses: docker/setup-buildx-action@v3.11.1
 
       - name: Login to GitHub Container Registry
-        uses: docker/login-action@v3
+        uses: docker/login-action@v3.6.0
         with:
           registry: ${{ secrets.DOCKER_REGISTRY_URL }}
           username: ${{ secrets.DOCKER_REGISTRY_USERNAME }}
@@ -28,7 +28,7 @@ jobs:
 
       - name: Docker metadata
         id: meta
-        uses: docker/metadata-action@v5
+        uses: docker/metadata-action@v5.8.0
         with:
           images: ${{ secrets.DOCKER_REGISTRY_URL }}/hyperion
           tags: |
@@ -38,7 +38,7 @@ jobs:
             preprod.branch=${{ github.event.inputs.branch }}
 
       - name: Build and push preprod image
-        uses: docker/build-push-action@v6
+        uses: docker/build-push-action@v6.18.0
         with:
           context: .
           platforms: linux/amd64 #,linux/arm64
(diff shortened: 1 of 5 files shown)


requirements-prod.txt
```diff
@@ -1,2 +1,2 @@
 -r requirements-common.txt
-psycopg[c]==3.1.19                  # PostgreSQL adapter for synchronous operations at startup (database initializations & migrations), local installation is recommended for a production site
\ No newline at end of file
+psycopg[binary]==3.1.19                  # PostgreSQL adapter for synchronous operations at startup (database initializations & migrations), local installation is recommended for a production site
```
armanddidierjean, line 2: If we want to use the binary version, this should become a common requirement

Dockerfile
```diff
@@ -1,18 +1,46 @@
...
-FROM ${DOCKER_REGISTRY_IDENTIFER}/hyperion-base:${REQUIREMENTS_MD5}
+FROM ghcr.io/astral-sh/uv:python3.12-trixie-slim
 
+# Default number of workers; can be overridden at runtime
+ENV WORKERS=1
+
+# Update package list and install weasyprint dependencies
+RUN apt-get update && apt-get install -y \
+    weasyprint \
+    && rm -rf /var/lib/apt/lists/*
```
armanddidierjean, line 9: Could you add comments explaining why we install weasyprint manually, and why we need to delete a folder?

Dockerfile
```diff
@@ -1,18 +1,46 @@
...
+
+# Install dependencies using uv (beaucoup plus rapide)
+RUN uv pip install --system --no-cache -r requirements-prod.txt
+
+# Then copy the rest of the application code
 COPY pyproject.toml .
 COPY alembic.ini .
 COPY migrations migrations/
 COPY assets assets/
 COPY app app/
```
armanddidierjean, line 34: I think some of these copy could be done earlier, as the files are less likely to change than requirements: alembic.ini and assets for example

Dockerfile
```diff
@@ -1,18 +1,46 @@
...
+# Change ownership of the application directory to the hyperion user
+RUN chown -R hyperion:hyperion /hyperion
+
+# Switch to non-root user
+USER hyperion
+
+# Expose port 8000
+EXPOSE 8000
+
+# Use uvicorn for better performance and flexibility
```
armanddidierjean, line 45: In what way does uvicorn give better performance and flexibility than fastapi cli?

Dockerfile
warix8: I feel that it is better to copy them after packages installation to use the cache for pip install even if there is a migration for instance. What would be the point of putting them first ?

Dockerfile
warix8: There is no real difference since fastapi will use uvicorn. I thought it would remove a layer of abstraction, but perhaps it actually makes understanding more complex.

PR #876, opened 2025-10-14
Remove `requests`, use `httpx` for sync and async HTTP requests

Marc-Andrieu:
### Summary

Split #845 into a 3rd PR with only the removal of `requests` and its major consequence: send logs asynchronously.

## Changes Made

- replace all `requests` use with `httpx`, or remove it
  - consequence: the asynchronicity of sending logs propagates, and is escaped using FastAPI's `BackgroundTasks`

## Type of Change

## Impact & Scope

- [x] Core functionality changes
- [ ] Single module changes
- [ ] Other

## Testing

- [x] Added/modified tests that pass the CI
- [ ] Tested in a pre-prod
- [x] Tested this locally

## Documentation

- [ ] Updated docs accordingly (docs.myecl.fr) : 
- [ ] Code includes docstrings
- [x] No documentation needed

### DIFF
--- a/app/utils/loggers_tools/matrix_handler.py
+++ b/app/utils/loggers_tools/matrix_handler.py
@@ -20,6 +20,7 @@ class MatrixHandler(StreamHandler):
 
     def __init__(
         self,
+        background_tasks: BackgroundTasks,
         room_id: str,
         token: str,
         server_base_url: str | None,
@@ -29,6 +30,7 @@ class MatrixHandler(StreamHandler):
         super().__init__()
         self.setLevel(level)
 
+        self.background_tasks = background_tasks
         self.room_id = room_id
         self.enabled = enabled
         if self.enabled:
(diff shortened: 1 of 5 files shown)


app/utils/loggers_tools/matrix_handler.py
```diff
@@ -42,9 +45,12 @@ def __init__(
     def emit(self, record):
         if self.enabled:
             msg = self.format(record)
-
             try:
-                self.matrix.send_message(self.room_id, msg)
+                self.background_tasks.add_task(
```
armanddidierjean: Loggers are encapsulated in a `QueueHandlers` which send logs in a dedicated thread. I'm honstly unsure about the consequences of using FastAPI BackgroundTask here.
I believe:
 - it should be unneeded as logs are already sent in their own thread
 - depending on BackgroundTask implementation, it could pull back the log emission in the main thread, which would be counterproductive
 - the execution of the task would be planned by FastAPI, at an unknown instant. We have no guarantee that a task added outside endpoint executions would be instantly scheduled.

I these fears are unfounded, I think it would be a good thing to add comments in the code explaining the benefits of using a BackgroundTask here

Marc-Andrieu: You're right... Even the native logging HTTPHandler (https://docs.python.org/3/library/logging.handlers.html#httphandler) seems synchronous...

After all I don't care much about the "sync vs async" debate *per se*, I just want:
- to remove `requests` and `type_requests` from our dependencies
- not to alter the current performance

Marc-Andrieu: Update: this PR is about to become a LOT easier!
The very 1st sentence of https://www.python-httpx.org/ states that `httpx` "*provides sync and async APIs*"

armanddidierjean (approved):

warix8 (approved):

PR #896, opened 2025-11-10
MyECLPay's 2nd payment method: the "classical" purchase

Marc-Andrieu:
# Description

## Summary

As of now, MyECLPay has only one payment method: the QR Code scanning, involving a seller user who makes the API call.
This PR adds a second payment method: the "classical" purchase, involving only the buyer, who makes the API call themself.

This 2nd payment method is meant to be used by other modules: AMAP, purchases itself, web front-ends indirectly (CdR, ...), etc.

## Changes Made

- [x] Aded endpoint to make a purchase at the buyer user's request, with unforgiving business logic
- [x] Refacto stuff in common with the scan payment method: signature verification, `TransactionRequestInfo` model (the one with the total amount, the IAT, the signature etc), the `UsedTransactionRequest` table for idempotency (can't be debited twice if retrying),
- [x] New `INDIRECT` transaction type (because distant, as opposed to `DIRECT` for scanning which involves a direct seller user). Also the history types has new `INDIRECT_GIVEN` and `INDIRECT_RECEIVED` (the direct transaction type is still associated to `GIVEN` and `RECEIVED`), I'm not sure of this design however!
- [x] A 20 second expiration: perhaps too short? (the idea: if you have a slow network and ragequit, you won't be debited mysteriously due to some retry in the network)
- [ ] Migration and tests

## Additional Notes

Differences with the scan payment method:
- No notion of membership check endpoint and bypass: you have to have the membership (if the store's structure requires one)
- The model has no `store` bool: because the path is `/myeclpay/stores/{store_id}/purchase`, you can only pay a store by design
- We verify the user performing the request is indeed the owner of the device that signed the transaction request

<details>
<summary>

# Classification

</summary>

## Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)
- [ ] 😶‍🌫️ No impact for the end-users

## Impact & Scope

- [x] Single module changes
- [x] Database migrations required
- [ ] Other: ... 

## Testing

- [ ] 1. Tested this locally

## Documentation

- [x] `"` Docstrings
- [x] `#` Inline comments
- [ ] No documentation needed

</details>

### DIFF
--- a/app/core/mypayment/cruds_mypayment.py
+++ b/app/core/mypayment/cruds_mypayment.py
@@ -999,6 +999,42 @@ async def delete_used_qrcode(
     )
 
 
+async def create_used_payment(
+    payment: schemas_myeclpay.PurchaseInfo,
+    db: AsyncSession,
+) -> None:
+    wallet = models_myeclpay.UsedPurchase(
+        payment_id=payment.id,
+        payment_tot=payment.tot,
+        payment_iat=payment.iat,
+        payment_key=payment.key,
+        signature=payment.signature,
+    )
+    db.add(wallet)
+
+
+async def get_used_payment(
+    payment_id: UUID,
+    db: AsyncSession,
+) -> models_myeclpay.UsedPurchase | None:
+    result = await db.execute(
+        select(models_myeclpay.UsedPurchase).where(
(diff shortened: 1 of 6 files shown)


Marc-Andrieu:
Superseded by #982

PR #913, opened 2025-11-29
Flappybird fix

cotanoine:
Some data was stored in db but never accessed


Rotheem (approved):

Marc-Andrieu (approved):

Marc-Andrieu:
But don't forget to update the migration file (filename + down_revision)

PR #916, opened 2025-11-30
Simplify workers logic.

julien4215:
Simplify workers logic.

If there is no redis database, it is assumed that there is no worker running. Otherwise redis lock is used.


Marc-Andrieu (commented):
wtf?

PR #985, opened 2026-04-12
Async files manipulation using anyio

armanddidierjean:
# Description

## Summary

- Starlette and FastAPI are based on [anyio](https://anyio.readthedocs.io/en/stable/), see [FastAPI documentation](https://fastapi.tiangolo.com/async/#write-your-own-async-code)
- use `anyio.Path` instead of `pathlib.Path`, see [ASYNC240](https://docs.astral.sh/ruff/rules/blocking-path-method-in-async-function/)
- remove `aiofiles` dependency in favor of `anyio`

<details>
<summary>

# Classification

</summary>

## Type of Change

## Impact & Scope

- [ ] Single module changes
- [ ] Other: ... 

## Testing

- [ ] 1. Tested this locally

## Documentation

- [ ] `"` Docstrings
- [ ] `#` Inline comments
- [x] No documentation needed

</details>


warix8:
Could you please update https://docs.myecl.fr/hyperion/other/data_files.html accordingly ?

armanddidierjean:
> Could you please update https://docs.myecl.fr/hyperion/other/data_files.html accordingly ?

See https://github.com/aeecleclair/myecl-documentation/pull/15

warix8 (approved):

app/core/utils/config.py
```diff
@@ -1,5 +1,5 @@
+import pathlib
```
Toshyb13, line 1: Why is this one with pathlib and not anyio like the others?

armanddidierjean: It seems it is not possible to create an async *property*, according to https://github.com/pydantic/pydantic/discussions/11509

pyproject.toml will only be read once per worker, at startup, the impact of using synchronous methods instead of asynchronous variant should be limited.

armanddidierjean: This is a compromize, there could maybe be better ways to implement this, if one as a better idea

NakoGH (approved):

NakoGH (approved):

PR #1010, opened 2026-05-27
Timeout between each email recovery

Bombefire:
# **Summary :**

We implement a 5-minute cooldown between password reset requests. The CRUD layer now checks if a reset request was made for a given email within the last 5 minutes. If a recent request exists, the endpoint returns an error. Otherwise, it proceeds normally with token generation and email sending. The 5 min can be changed in the settings.

We handle two groups : User with and without an account. For user with an account everything is set and we just verify the 5 minutes mark. For the others, we create a new table to stock their recovery attempts and impose in this way the 5 minutes cooldown. We log their first attempt to log in a seperate table with the time and email used.


app/core/users/cruds_users.py
```diff
@@ -137,7 +138,20 @@ async def get_user_by_email(
     )
     return result.scalars().first()
 
-
+async def get_last_recovory_demand_last_time(
```
Daihecyy: Recovery

app/core/users/cruds_users.py
```diff
@@ -137,7 +138,20 @@ async def get_user_by_email(
...
+    db: AsyncSession,   
+    email: str,
+    time: datetime,
```
Daihecyy: Pk tu mets un datetime ? C'est pas plutôt un float nombre de minutes ?

Daihecyy: C'est toujours bien d'être explicite sur les unités,  genre minimumDelayMinutes

app/core/users/endpoints_users.py
```diff
@@ -607,6 +613,11 @@ async def recover_user(
             hyperion_security_logger.info(
                 f"Reset password for {email}: {calypsso_reset_url}",
             )
+    else:
```
Daihecyy, line 654: Plutôt que de passer par un else, tu pourrais return early avec ton erreur

[https://en.wikipedia.org/wiki/Cyclomatic_complexity]()

app/core/utils/config.py
```diff
@@ -342,6 +342,7 @@ def settings_customise_sources(
 
     USER_ACTIVATION_TOKEN_EXPIRE_HOURS: int = 24
     PASSWORD_RESET_TOKEN_EXPIRE_HOURS: int = 12
+    PASWORD_RECOVERY_NEW_TOKEN_MINUTES: int = 5
```
Daihecyy: :pick: NEW_TOKEN_EXPIRE_MINUTES ?

tests/core/test_users.py
Daihecyy, line 1: Là ça va parce que le delay est long mais c'est une recette pour avoir des tests flaky de se baser sur le temps réel 
Si ta CI change de VM en plein milieu ton test peut fail sans raison

app/core/users/cruds_users.py
Bombefire: C'est un int

Bombefire: Pas faux

tests/core/test_users.py
Bombefire: Comment faire pour résoudre le problème vue qu'on doit vérifier que 5min s'est écoulé ?

Daihecyy: Dans le cas présent c'est pas hyper grave
C'est un peu chiant à faire et je sais pas comment on le ferait avec FastApi mais tu veux mocker datetime pour avoir un comportement deterministique

app/core/users/cruds_users.py
```diff
@@ -195,6 +243,15 @@ async def create_user(
     return user
 
 
+async def create_unregistred_user(
+    db: AsyncSession,
+    user: models_users.CoreUnregistredUser,
+) -> models_users.CoreUnregistredUser:
+    db.add(user)
+    await db.flush()
```
armanddidierjean: Do you really need to flush?

app/core/users/cruds_users.py
```diff
@@ -421,3 +478,12 @@ async def fusion_users(
...
     # Delete the user_deleted
     await delete_user(db, user_deleted_id)
+
+
+async def create_unregistred_user_recover_request(
+    db: AsyncSession,
+    recover_request: models_users.CoreUnregistredUserRecoverRequest,
+) -> models_users.CoreUnregistredUserRecoverRequest:
+    db.add(recover_request)
+    await db.flush()
```
armanddidierjean: Do you really need to flush?

app/core/users/endpoints_users.py
```diff
@@ -605,9 +600,78 @@ async def recover_user(
             )
         else:
             hyperion_security_logger.info(
-                f"Reset password for {email}: {calypsso_reset_url}",
+                f"Reset password failed for {email} due to not active function",
```
armanddidierjean: The idea was to log the reset_url: during development, you won't have an enabled smtp server. It is thus usefull to be able to fetch the url from hyperion logs


```suggestion
                f"Reset password for {email}: {calypsso_reset_url}",
```

app/core/users/endpoints_users.py
```diff
@@ -605,9 +600,78 @@ async def recover_user(
...
+        )
+        send_email(
+            recipient=email,
+            subject="MyECL - reset your password",
+            content=mail,
+            settings=settings,
+        )
+    else:
+        hyperion_security_logger.info(
+            f"Reset password failed for {email} due to not active function",
```
armanddidierjean: It would probably be clearer to indicate precisely why the request failed
```suggestion
            f"Reset password failed for {email}, user does not exist",
```

app/core/users/endpoints_users.py
```diff
@@ -547,39 +547,34 @@ async def recover_user(
...
-        else:
-            hyperion_security_logger.info(
-                f"Reset password failed for {email}, user does not exist",
-            )
+    reset_token = security.generate_token()
 
-    else:
+    if last_created is not None:
+        hyperion_security_logger.info(
+            "We are in cooldown",
```
armanddidierjean: Maybe a more explicit message could be better. Like: `Reset password failed: user {email} is in cooldown`

app/core/users/models_users.py
```diff
@@ -108,3 +108,22 @@ class CoreUserEmailMigrationCode(Base):
 
     # If the user should become an external or a member user after the email change
     make_user_external: Mapped[bool] = mapped_column(default=False)
+
+
+class CoreUnregistredUser(Base):
```
armanddidierjean: It may be easier to use only one table with a datetime:
```python
class PasswordRecoverForUnregistredUser:
    email: str
    last_recover_send: datetime
```

app/core/users/cruds_users.py
Bombefire: We don't need the user after this so their is no need to flush. (Can you expand in other specific example of using flush or not) ?

app/core/users/models_users.py
Bombefire: Staying consistent with the code before

app/core/users/cruds_users.py
```diff
@@ -138,6 +139,68 @@ async def get_user_by_email(
     return result.scalars().first()
 
 
+async def get_recovery_request_within_delay_registred_user(
```
Daihecyy, line 142: :pick: typo, c'est registered (Idem partout où il est utilisé)

app/core/users/endpoints_users.py
```diff
@@ -608,6 +602,61 @@ async def recover_user(
                 f"Reset password for {email}: {calypsso_reset_url}",
             )
 
+        return standard_responses.Result()
+
+    # We check now if this unregistred mail is under cooldown
+    # We check now if this unregistred mail is under cooldown
```
Daihecyy: :pick: duplicate line

app/core/users/cruds_users.py
Bombefire: ?

app/core/users/models_users.py
```diff
@@ -108,3 +108,18 @@ class CoreUserEmailMigrationCode(Base):
...
     # If the user should become an external or a member user after the email change
     make_user_external: Mapped[bool] = mapped_column(default=False)
+
+
+class CoreUnregistredUserRecoverRequest(Base):
+    __tablename__ = "core_unregistred_user_recover_request"
+
+    # The email column should not be unique.
+    # Someone can indeed create more than one password reset request,
+    id: Mapped[int] = mapped_column(
```
armanddidierjean: Please use a `PrimaryKey` (an uuid) instead of int

app/core/users/cruds_users.py
```diff
@@ -138,6 +139,68 @@ async def get_user_by_email(
...
+    db: AsyncSession,
+    email: str,
+    minimumDelayMinutes: int,
```
armanddidierjean: We usually use snake case
```suggestion
    minimum_delay_minutes: int,
```
You could also directly pass a timedelta, if you prefer, but an int is fine too

app/core/users/cruds_users.py
```diff
@@ -138,6 +139,68 @@ async def get_user_by_email(
...
+        )
+        if result
+        else None
+    )
+
+
+async def get_recovery_request_within_delay_unregistred_user(
+    db: AsyncSession,
+    email: str,
+    minimumDelayMinutes: int,
```
armanddidierjean: We usually use snake case

app/core/utils/config.py
```diff
@@ -342,6 +342,7 @@ def settings_customise_sources(
 
     USER_ACTIVATION_TOKEN_EXPIRE_HOURS: int = 24
     PASSWORD_RESET_TOKEN_EXPIRE_HOURS: int = 12
+    PASWORD_RECOVERY_NEW_TOKEN_EXPIRE_MINUTES: int = 5
```
armanddidierjean: Typo

```suggestion
    PASSWORD_RECOVERY_NEW_TOKEN_EXPIRE_MINUTES: int = 5
```

armanddidierjean: I may have misunderstood the purpose of this variable, but shouldn't it be something like `PASSWORD_RECOVERY_TOKEN_COOLDOWN_MINUTES` instead? It doesn't seem that the token expires after 5 minutes.
Maybe a comment explaining what this settings does would also be useful

armanddidierjean (dismissed):

migrations/versions/59-unregistered_recovery_requests.py
Marc-Andrieu, line 1: 3 things : 
- rename it for consistency: migration files begin indeed with the n° but having - or _ between words is better
- do not forget to fix the missing "e" in "registered" the table name here as well
- This id column is useless (notice you never actually use it in CRUDs) : the email+created_on pair *is* the real pkey (just so you start here with good practices 😉)

Bombefire: If a person resend two times a recovery request we can have two mail on different lines. If two person send at the same time a recovery request we have the same created on on different lines. So we should have a an id of each request.

armanddidierjean: You could delete the previous one, or update the datetime. It is not really useful to store passed requests, only the most recent will be used

app/core/users/endpoints_users.py
```diff
@@ -608,6 +602,61 @@ async def recover_user(
...
+        )
+        send_email(
+            recipient=email,
+            subject="MyECL - reset your password",
+            content=mail,
+            settings=settings,
+        )
+    else:
+        hyperion_security_logger.info(
+            f"Reset password for {email} due to inactive option",
```
armanddidierjean, line 656: What do you mean by "inactive option"?

migrations/versions/59-unregistered_recovery_requests.py
Luhmos: > If a person resend two times a recovery request we can have two mail on different lines. If two person send at the same time a recovery request we have the same created on on different lines. So we should have a an id of each request.

What Marc-Andrieu meant was that you can use the pair of email+created_on *together* as your key, not separately.

app/core/utils/config.py
Bombefire: Have been changed to PASSWORD_RECOVERY_TOKEN_COOLDOWN_MINUTES plus comment for more clarity

tests/core/test_users.py
Bombefire: Changed test to use moker for datetime

app/core/users/endpoints_users.py
Bombefire: Changed for users with active accounts but for unregistred users it wasn't changed

app/core/users/cruds_users.py
Bombefire: Didn't see the typo now changed accordingly

app/core/users/endpoints_users.py
Bombefire: Deleted the duplicated line

migrations/versions/59-unregistered_recovery_requests.py
Bombefire: We are know using two colomns as primary keys. We dropped the use of an id as it isn't used.

app/core/users/endpoints_users.py
Bombefire: The option to create an account is bound by SMTP_ACTIVE so we can't send a a mail to unregistred users (a mail to create an account).

app/core/users/endpoints_users.py
Bombefire: Changed

NakoGH (approved):
Great first PR 👍  ! Keep going !!!

app/core/users/endpoints_users.py
```diff
@@ -547,40 +547,34 @@ async def recover_user(
...
         recover_request = models_users.CoreUserRecoverRequest(
             email=email,
             user_id=db_user.id,
             reset_token=reset_token,
-            created_on=datetime.now(UTC),
-            expire_on=datetime.now(UTC)
-            + timedelta(hours=settings.PASSWORD_RESET_TOKEN_EXPIRE_HOURS),
+            created_on=date,
+            expire_on=date
+            + timedelta(hours=settings.PASSWORD_RECOVERY_TOKEN_COOLDOWN_MINUTES),
```
Toshyb13: Souldn't the allowed use of the reset link be different from the delay between two requests?

Toshyb13:
```suggestion
            + timedelta(hours=settings.PASSWORD_RESET_TOKEN_EXPIRE_HOURS),
```

Toshyb13 (dismissed):

app/core/users/endpoints_users.py
Bombefire: Thank you for you review. I changed the code now.

app/core/users/cruds_users.py
warix8: registred -> registered
typo is still there?

app/core/users/models_users.py
```diff
@@ -108,3 +108,12 @@ class CoreUserEmailMigrationCode(Base):
...
     make_user_external: Mapped[bool] = mapped_column(default=False)
+
+
+class CoreUnregisteredUserRecoverRequest(Base):
+    __tablename__ = "core_unregistered_user_recover_request"
+
+    # The email column should not be unique.
+    # Someone can indeed create more than one password reset request,
+    email: Mapped[str] = mapped_column(primary_key=True)
+    created_on: Mapped[datetime] = mapped_column(primary_key=True)
```
warix8, line 119: Why created_on column is a primary key ?

app/core/users/models_users.py
```diff
@@ -108,3 +108,12 @@ class CoreUserEmailMigrationCode(Base):
 
     # If the user should become an external or a member user after the email change
     make_user_external: Mapped[bool] = mapped_column(default=False)
+
+
+class CoreUnregisteredUserRecoverRequest(Base):
+    __tablename__ = "core_unregistered_user_recover_request"
+
+    # The email column should not be unique.
+    # Someone can indeed create more than one password reset request,
```
warix8, line 117: For unregistered user recover request, it should be unique, no need of multiple refresh tokens, just store email -> date to check the cooldown

app/core/users/endpoints_users.py
```diff
@@ -608,6 +602,60 @@ async def recover_user(
...
+
+    recover_request_unregistered = models_users.CoreUnregisteredUserRecoverRequest(
+        email=email,
+        created_on=date,
+    )
+
+    await cruds_users.create_unregistered_user_recover_request(
+        db=db,
+        recover_request=recover_request_unregistered,
+    )
```
warix8, line 635: Create or update the existing record, we could store only one row for each email in the table. Moreover, you have set column email as a primary key, so it shouldn't be possible to create more than one record (for each attempt not on cooldown) for unregistered db table.

warix8 (commented):
A very good first PR.

app/core/users/models_users.py
Bombefire: A person can make more than one recovery request so it can't be a primary key. And if a two person make at the same time a recovery request they will have the same created_on.

app/core/users/models_users.py
Bombefire: Their is no token for unregistred users

warix8: Yes, I know but since there are no tokens, there is no need to have multiple rows for an email of an unregistered user. We can only update the date of the last request for this specific email. One row per email as suggested by @armanddidierjean.

app/core/users/endpoints_users.py
Bombefire: I don't think that will optimize updating than recreating each time a new recovery request.

Toshyb13 (approved):

app/core/users/endpoints_users.py
warix8: I'll leave it up to you, it was just a suggestion
But you have to decide whether the email address or the date is the primary key, not both.
Currently in your model
```py
email: Mapped[str] = mapped_column(primary_key=True)
created_on: Mapped[datetime] = mapped_column(primary_key=True)
```

PR #1012, opened 2026-06-04
Run tests in parallel

NakoGH:
# Description

## Summary

Test suite was running all modules serially (> 3min), with profiling showing that 100% of the overhead was in the setup phase (~3–5s per module × 35 modules). This PR runs test modules in parallel using pytest-xdist which brings total runtime to ~60s.

A problem occured with isolation: each xdist worker (gw0, gw1, …) gets its own dedicated PostgreSQL database (hyperion_gw0, etc.) created at session start and dropped at teardown, so parallel workers never share state. 

A secondary issue is that `get_number_of_workers()` in `initialization.py` uses `psutil` to count children of the parent process — with xdist that counts the other workers rather than app workers — causing `use_lock_for_workers` to skip `init_db` on all but one worker. This is patched/tricked in the client fixture.

...

## Changes Made

- [x] add `pytest-xdist` to requirements-dev.txt
- [x] modified default tests conf  to run tests with 8 workers
- [x]  

## Additional Notes

<details>
<summary>

# Classification

</summary>

## Type of Change

- [ ] 🔨 Refactor (non-breaking change that neither fixes a bug nor adds a feature)
- [x] 🔧 Infra CI/CD (changes to configs of workflows)

## Impact & Scope

- [ ] Single module changes
- [x] Other: ... 

## Testing

- [x] 1. Tested this locally

## Documentation

- [ ] `"` Docstrings
- [ ] `#` Inline comments
- [x] No documentation needed

</details>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -175,6 +175,7 @@ asyncio_mode = "auto"
 asyncio_default_fixture_loop_scope = "module"
 asyncio_default_test_loop_scope = "module"
 filterwarnings = ["error", "ignore::DeprecationWarning"]
+addopts = "-n 8 --dist=loadfile" # loadfile makes sure that tests are grouped by their containing file and distributed to available workers as whole units.
 
 
 [tool.coverage.run]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -174,7 +174,13 @@ ignore_missing_imports = true
 asyncio_mode = "auto"
 asyncio_default_fixture_loop_scope = "module"
 asyncio_default_test_loop_scope = "module"
-filterwarnings = ["error", "ignore::DeprecationWarning"]
+filterwarnings = [
+    "error",
+    "ignore::DeprecationWarning",
+    # xdist controller imports `app` during collection before per-worker coverage starts;
+    # the warning is benign -- workers still measure correctly.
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -175,6 +175,7 @@ asyncio_mode = "auto"
 asyncio_default_fixture_loop_scope = "module"
 asyncio_default_test_loop_scope = "module"
 filterwarnings = ["error", "ignore::DeprecationWarning"]
+addopts = "-n 8 --dist=loadfile" # loadfile makes sure that tests are grouped by their containing file and distributed to available workers as whole units.
```
Luhmos, line 184: Why n=8? Why not get the current cpu's information, or let the user pick or something else?

NakoGH: It can be overridden by anyone running ``` pytest -n nb_workers``` in the terminal. We could use the cpu info using ``` -n auto``` but we will get limited improvements on CI because github runners only have 2 cores. Howerver, the os is able to schedule multiple python process (workers) and we will get better results here because tests  spend time mainly on I/O (postgres db) and not on pure logic.  

Still, the choice of 8 workers is a bit arbitrary, it fast enough to run locally and we get better results on Ci but we need to check/test the results running the jobs.

Toshyb13 (approved):

Rotheem (approved):

PR #1054, opened 2026-08-24
Raid v2 improvments

maximeroucher:
# Description

## Summary

Updating Raid to match the first points for data security

## Changes Made

- [x] Adding consent field to the security file
- [x] Adding read_medical_data permission gating the zip download (now only source of security file read)
- [x] Logging when a zip download is attempted 
- [x] Adding course responsible as emergency contact into the security file, as the Raid requested

## Additional Notes

First step toward a better security data handling.

<details>
<summary>

# Classification

</summary>

## Type of Change

- [ ] 🔨 Refactor (non-breaking change that neither fixes a bug nor adds a feature)

## Impact & Scope

- [x] Single module changes
- [x] Database migrations required
- [ ] Other: ... 

## Testing

- [x] 1. Tested this locally

## Documentation

- [ ] `"` Docstrings
- [x] `#` Inline comments
- [ ] No documentation needed

</details>


app/modules/raid/coredata_raid.py
```diff
@@ -13,6 +13,7 @@ class RaidInformation(core_data.BaseCoreData):
     president: EmergencyContact | None = None
     volunteer_responsible: EmergencyContact | None = None
     security_responsible: EmergencyContact | None = None
+    course_responsible: EmergencyContact | None = None
```
Marc-Andrieu, line 16: Not sure you want to keep French "course" but ok

app/modules/raid/endpoints_raid.py
```diff
@@ -740,24 +752,17 @@ async def set_security_file(
...
-            edition.id,
-            db,
-        )
-        target_team = await cruds_raid.get_team_by_participant_id(
-            participant_id,
-            edition.id,
-            db,
-        )
-        if user_team is None or target_team is None or user_team.id != target_team.id:
-            raise HTTPException(status_code=403, detail="You are not the participant.")
```
Marc-Andrieu, line 757: Oh, I didn't know that until now you could change your teammate's file

Marc-Andrieu (approved):
Looks promising! That's a clean start about the medical data issue

app/modules/raid/coredata_raid.py
maximeroucher: It is an english term

app/modules/raid/endpoints_raid.py
maximeroucher: You were supposed to register your team mate without him/her to use the app

app/modules/raid/coredata_raid.py
Marc-Andrieu: k

app/modules/raid/cruds_raid.py
```diff
@@ -58,10 +58,16 @@ async def get_all_participants(
...
-    ]
+
+    # Remove security_file from the participants list to avoid including it in the response.
+    found_participants = participants.scalars().all()
+    cleaned_participants = []
+    for p in found_participants:
+        participant = schemas_raid.RaidParticipant.model_validate(p)
+        participant.security_file = None
+        cleaned_participants.append(participant)
+
```
Daihecyy, line 65: ⚠️ you do a `selectinload("*")` right above, narrowing it could remove this ?
Here you unnecessarily load data in your server (bad for performance) and the data is sensitive (bad for security)

app/modules/raid/endpoints_raid.py
```diff
@@ -202,16 +204,26 @@ async def get_participant_by_id(
     ),
     edition: schemas_raid.RaidEdition = Depends(get_current_raid_edition),
 ):
-    if user_id != user.id and not await has_user_permission(
+    is_owner = user.id == user_id
+    is_admin = await has_user_permission(
```
Daihecyy: ⛏️ `is_raid_admin` ?

app/modules/raid/endpoints_raid.py
```diff
@@ -202,16 +204,26 @@ async def get_participant_by_id(
...
             status_code=403,
             detail="You can not get data of another user",
         )
-    return await get_participant_or_404(user_id, edition.id, db)
+
+    participant = await get_participant_or_404(user_id, edition.id, db)
+
+    # If the user is not the owner and does not have medical permission, hide the security file
+    if not is_owner:
+        participant.security_file = None
```
Daihecyy: 💡 Here again, would really benefit from a narrower `selectinload` but would require more substantial work

app/modules/raid/endpoints_raid.py
```diff
@@ -202,16 +204,26 @@ async def get_participant_by_id(
...
+    if not is_owner and not is_admin:
         raise HTTPException(
             status_code=403,
             detail="You can not get data of another user",
         )
-    return await get_participant_or_404(user_id, edition.id, db)
+
+    participant = await get_participant_or_404(user_id, edition.id, db)
+
+    # If the user is not the owner and does not have medical permission, hide the security file
```
Daihecyy: ❓ Comment disagrees with code, which one do we want ?

app/modules/raid/endpoints_raid.py
```diff
@@ -740,24 +752,17 @@ async def set_security_file(
...
+        raise HTTPException(status_code=403, detail="You are not the participant.")
 
     participant = await get_participant_or_404(participant_id, edition.id, db)
 
+    if not security_file.consent_given:
+        raise HTTPException(
+            status_code=400,
+            detail="Consent must be given to register medical data",
+        )
+
```
Daihecyy, line 837: ❓ Could replace this with `consent_given: Literal[True]` in the schema ?

app/modules/raid/endpoints_raid.py
```diff
@@ -1158,8 +1164,30 @@ async def download_security_files_zip(
...
+    has_medical_permission = await has_user_permission(
+        user,
+        RaidPermissions.read_medical_data,
+        db,
+    )
+
+    if not has_medical_permission:
+        raise HTTPException(
+            status_code=403,
+            detail="You can not get data of another user",
```
Daihecyy: ⛏️ Misleading comment, it is _all users_

app/modules/raid/endpoints_raid.py
```diff
@@ -1158,8 +1164,30 @@ async def download_security_files_zip(
...
     zip_file_path = await get_all_security_files_zip(db, information, edition.id)
+
+    hyperion_security_logger.info(
+        "Medical data access",
+        extra={
+            "accessed_by_user_id": user.id,
+            "edition_id": str(edition.id),
+            "access_type": "download",
+        },
+    )
```
Daihecyy, line 1343: 💡 Archi wise, `get_all_security_files_zip` should own the logging for me, not the endpoint, it is the one doing the critical step
This way, we futureproof it and won't forget to log elsewhere
I would also have put the permission check in it

app/modules/raid/models_raid.py
```diff
@@ -72,6 +72,7 @@ class SecurityFile(Base):
     emergency_person_name: Mapped[str | None]
     emergency_person_phone: Mapped[str | None]
     file_id: Mapped[str | None]
+    consent_given: Mapped[bool] = mapped_column(default=False)
```
Daihecyy, line 75: 🛑  `consent_given`  as a boolean doesn't represent a consent for me
First, it is mandatory to fill in the mandatory security sheet so not free at all (in `set_security_file`)
Second, we could have wanted a timestamp and/or a version number of the consent text. Here, we don't know what the user consented to nor when
Retractation doesn't seem to be handled that well, especially considering we have no historization

Daihecyy (commented):
Possible to split PR ?
`consent_given` and `course_responsible` have nothing to do with each other, comments will get muddled

app/modules/raid/models_raid.py
Marc-Andrieu: Something along the lines of MyPayment
https://github.com/aeecleclair/Hyperion/blob/acba6e2/app/core/mypayment/models_mypayment.py#L244

app/modules/raid/endpoints_raid.py
Marc-Andrieu: ?

app/modules/raid/models_raid.py
Daihecyy: It doesn't handle retractation either but it would be a step in the right direction

app/modules/raid/endpoints_raid.py
Marc-Andrieu: You mean, changing the CRUD so that its behavior depends on the end-user and the SQL may return null?

Daihecyy: No, simply passing a boolean param (`with_security_file`) or duplicating the CRUD with a medically blind version so that security_file is never instantiated (and calling the right function depending on the conditions)

maximeroucher:
> Possible to split PR ? `consent_given` and `course_responsible` have nothing to do with each other, comments will get muddled

Given the time constraint and the time taken to review and merge a PR, I decided to put all the changes in one. I acknowledge that it will be a bit harder to read, that's why I ensure to produce really thin commit to allow a good review

app/modules/raid/cruds_raid.py
maximeroucher: Agreed, do you know how to do it better, I still want to pull the rest of the data ?

app/modules/raid/endpoints_raid.py
maximeroucher: Fair

app/modules/raid/endpoints_raid.py
maximeroucher: Mb, earlier design choices, updated

app/modules/raid/endpoints_raid.py
maximeroucher: Not sure to get it either

app/modules/raid/endpoints_raid.py
maximeroucher: Fair

app/modules/raid/models_raid.py
maximeroucher: What do we do if a user rectract after the data are exported ?

app/modules/raid/endpoints_raid.py
maximeroucher: I don't see any loggers in cruds, except those for a db error, so I stuck to that architecture, maybe need the opinion of someone with more knowledge on the backend, at least on my side

app/modules/raid/endpoints_raid.py
Daihecyy: You are doing schema validation in the endpoint
you could replace this snippet by tightening the schema validator

app/modules/raid/cruds_raid.py
Daihecyy: ? Just put the list of the tables you want ?

app/modules/raid/models_raid.py
maximeroucher: And does versioning suppose to had their respective text in db as well ? If we change the text during a registering, what does happen, do we just remove registered info because the participant did not consent with the latest version, or do we consider that consenting to one imply consenting to all the following ones ?

app/modules/raid/cruds_raid.py
maximeroucher: In the selectinload, right ? like selectinload("app.modules.raid.models_raid.Document, raid_document.id, participant, CoreUser") ?

app/modules/raid/endpoints_raid.py
maximeroucher: Same selectinload I guess then ?

app/modules/raid/models_raid.py
Daihecyy: Well, it seems like medical consent is a freaking pain in the ass, who would have thought ?

app/modules/raid/endpoints_raid.py
Daihecyy: At this point, why even bother to put it in the schema, just put it at `True` no matter what

armanddidierjean: I disagree with this point. Putting `Literal[True]` would return a 422 validation error. 
In this endpoint, a value False should be accepted (in term of schema validation). It should however return a "400 Consent must be given to register medical data", as it violates a business rules: we can not proceed given a falsy consent. When receving this error, the frontend will explain to the user why it can not proceed

app/modules/raid/cruds_raid.py
armanddidierjean: Like this: https://github.com/aeecleclair/Hyperion/blob/e9b5f538bcc0376abe7f6c0a64a14c2c116d7356/app/core/mypayment/cruds_mypayment.py#L692-L695

app/modules/raid/models_raid.py
armanddidierjean: I agree, storing the datetime the consent was given at would be a good addition. I think automatic retractation is out of scope of this PR. I propose we start with manual suppression of data (including the deletion of files that were downloaded by the RAID association) if someone ask ECLAIR/the RAID for its data to be removed

app/modules/raid/cruds_raid.py
maximeroucher: Ok, is the last commit going in the right direction ?

app/modules/raid/models_raid.py
maximeroucher: I added the date, how would you design the retractation feature ? a simple endpoint marking consent as false and clearing the data ?

app/modules/raid/endpoints_raid.py
maximeroucher: @armanddidierjean Any advice on this one will be welcomed

armanddidierjean (approved):

app/modules/raid/endpoints_raid.py
```diff
@@ -225,20 +225,7 @@ async def get_participant_by_id(
...
-        t_shirt_size=participant.t_shirt_size,
-        situation=participant.situation,
-        payment=participant.payment,
-        t_shirt_payment=participant.t_shirt_payment,
-        user=participant.user,
-        validation_progress=participant.validation_progress,
-        attestation_on_honour=participant.attestation_on_honour,
-        is_minor=participant.is_minor,
-    )
+    return schemas_raid.RaidParticipantRestrictedComplete(**participant.model_dump())
```
armanddidierjean, line 228: You should keep the explicite instanciation of all fields, otherwise mypy won't be able to detect potential issues

PR #1058, opened 2026-08-25
Use two TestClient fixtures, raising or not exceptions in tests

armanddidierjean:
# Description

## Summary
Use two TestClient fixtures, raising or not exceptions in tests instead of never raising exceptions.
We may want to not raise exception and instead return a proper 500 response to assert that an InternalServerError happens in specific situations But raising exception should be the usual default, as InternalServerError should rarely be expected to happen. Raising the exception allows it to be logged by pytest and make debugging easier

<details>
<summary>

# Classification

</summary>

## Type of Change

## Impact & Scope

- [ ] Single module changes
- [x] Other: ... 

## Testing

- [ ] 1. Tested this locally

## Documentation

- [ ] `"` Docstrings
- [ ] `#` Inline comments
- [ ] No documentation needed

</details>

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,6 +4,7 @@ from unittest.mock import patch
 
 import psycopg
 import pytest
+from fastapi import FastAPI
 from fastapi.testclient import TestClient
 
 from app.app import get_application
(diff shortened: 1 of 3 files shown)


warix8 (dismissed):

tests/conftest.py
```diff
@@ -110,3 +131,32 @@ async def test_example(client: TestClient):
         TestClient(test_app, raise_server_exceptions=False) as client,
```
warix8: Shouldn't it be `raise_server_exceptions=True` for the default TestClient ?

armanddidierjean: Oh it should, nice catch, thanks

armanddidierjean:
```suggestion
        TestClient(test_app, raise_server_exceptions=True) as client,
```

Rotheem (approved):

warix8 (approved):

